PR #7, opened 2020-03-20
Disable cookie domain

nkassis:


indico/config/config.py
```diff
@@ -16,13 +16,14 @@ class IndicoConfig:
...
     api_token: str = None
     verify_ssl: bool = True
-
+    _disable_cookie_domain: bool = False
+    
     def __init__(self, **kwargs):
         self.host: str = os.getenv("INDICO_HOST", "app.indico.io")
         self.protocol: str = os.getenv("INDICO_PROTOCOL", "https")
         self.serializer: str = os.getenv("INDICO_SERIALIZER", "msgpack")
         self.api_token_path: str = os.getenv("INDICO_API_TOKEN_PATH", Path.home())
```
pastap: do you need to add `self._disable_cookie_domain` here?

pastap: Are the tests for this client being run?

dzereski: Yeah, Astha is right, I don't see _disable_cookie_domain getting used in this class.

nkassis: No you don't becaause it gets set by the look on kwargs if passed in. 

I do run the test locally right now.

PR #9, opened 2020-03-24
Updating DocumentExtraction docs

Scott771:
Added the import statements to the main example (this had confused me initially), made language more consistent, added '(default)' to kwarg values where I knew them, removed deprecated features (n_threads, etc.)


jkgenser:
Is the recommendation for people to default to `simple` in all cases? One thing to consider is that `simple` only supports native pdfs and not scans. Therefore, if you have a non-native PDF but don't want to use the `detailed` config, then you'd need to copy simple config and set the `native_pdf` key to `False`

madisonmay:
@jkgenser yeah this is troublesome.  Simple really is rarely the right option.  Feels like we should expose a "standard" config or similar that operates similarly to prior OCR.

PR #15, opened 2020-03-25
Object Detection

jkgenser:
Adds the ability to create a dataset from a list of images in a directory accessible by the python client. First uploads images to rainbow, then creates a csv of storage urls. This csv is then passed as the dataset file.

Also adds ability to pass in `model_training_options`. Added wait timeouts for the request chain because I would get server-side timeouts for spamming the platform for dataset status.


indico/queries/documents.py
```diff
@@ -5,27 +5,7 @@
 
 from indico.client.request import RequestChain, GraphQLRequest, HTTPMethod, HTTPRequest
 from indico.types.jobs import Job
-
-
-class _UploadDocument(HTTPRequest):
```
jkgenser: This seems redundant since the identical code is in storage.py.

indico/http/client.py
```diff
@@ -91,7 +91,8 @@ def _make_request(self, method: str, path: str, headers: dict=None, _refreshed=F
         with self._handle_files(request_kwargs):
             response = getattr(self.request_session, method)(
                 f"{self.base_url}{path}", headers=headers, stream=True, verify=self.config.verify_ssl, **request_kwargs
-            )   
+            )
+            print(response.text)
```
jkgenser, line 95: will remove

nkassis, line 95: remove

indico/client/request.py
```diff
@@ -48,6 +49,17 @@ def process_response(self, response):
...
     previous: Any = None
+    
+    def __init__(self):
+        self.timeout = 0
 
     def requests(self):
         pass
+
+    def wait_request(self):
+        self.timeout += 0.5
```
nkassis, line 60: this needs a max. Also instead of putting this in request chain, you could just make a  Debouncer class so you don't need to worry about resetting it, you can just use one per call you want to do this in.

indico/queries/datasets.py
```diff
@@ -197,6 +231,19 @@ def __init__(self, files: List[str]):
...
+class UploadImages(UploadDocument):
+    """
+    Upload image files stored on a local filepath to the indico platform.
+    """
+
+    def process_response(self, uploaded_files: List[dict]) -> List[str]:
+        errors = [f["error"] for f in uploaded_files if f.get("error")]
+        if errors:
+            return IndicoInputError(error="\n".join(error for error in errors),)
+        urls = [URL_PREFIX + f["path"] for f in uploaded_files]
```
nkassis, line 243: I might be misunderstanding but isn't the url prefix already on the path of these objects or are they not coming back as urls like for documents?

indico/queries/documents.py
nkassis: Yes forgot to move it to that class after making a common one.

indico/queries/datasets.py
jkgenser: No, they don't have URL prefixes coming off of rainbow `/store`. For documents, the filemeta is passed back and then doctor creates a storage object from un-prefixed url. In this case, we add the URL so that it can be identified as such for dataset upload.

indico/http/client.py
jkgenser: fixed.

indico/client/request.py
jkgenser: Used a separate debouncer class.

tests/integration/queries/test_model_group.py
```diff
@@ -95,6 +130,37 @@ def test_predict(indico, airlines_dataset, airlines_model_group):
     assert len(job.result) == 1
 
 
+def get_storage_urls_from_fnames(client, image_fnames):
+    client = IndicoClient()
+    image_filenames = ["1.jpg", "2.jpg", "3.jpg", "4.jpg", "5.jpg"]
```
jkgenser, line 135: going to fix this so it's not hard coded in this function.

indico/client/request.py
```diff
@@ -51,3 +52,17 @@ class RequestChain:
 
     def requests(self):
         pass
+
+
+class Debouncer:
+    def __init__(self):
+        self.timeout = 0
+        self.max_timeout = 20
```
nkassis, line 60: I think 5 ok, like we we don't need to debounce that much

jkgenser: updated to 5.

PR #19, opened 2020-03-31
Fix issue when using _disable_cookie_domain and cookies are being reset

nkassis:


indico/queries/datasets.py
```diff
@@ -80,6 +80,8 @@ def __init__(self, id: int):
 
     def process_response(self, response) -> Dataset:
         response = super().process_response(response)
+        if not "dataset" in response or type(response) != dict:
```
pastap: is it better to use `type` or `isinstance`? also should the checks be reversed? first type, then `in`

nkassis: actually this is all wrong it should be response["dataset"] and yeah instance, I won't lie I've been coding in another language recently with this symantic ;p

PR #20, opened 2020-04-02
Don/docs

dzereski:
Small changes to DocumentExtraction settings requested by Slater

### DIFF
--- a/docsrc/docextract_settings.rst
+++ b/docsrc/docextract_settings.rst
@@ -296,7 +296,7 @@ Setting::
 
 Return style information for the token. Example Return::
 
-    {"bold": true, "underlined": true, "italics": true, "font_size": 14, "background_color": "hex", "text_color": "hex"}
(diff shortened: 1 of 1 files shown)


docsrc/docextract_settings.rst
```diff
@@ -298,7 +296,7 @@ Setting::
 
 Return style information for the token. Example Return::
 
-    {"bold": true, "underlined": true, "italics": true, "font_size": 14, "background_color": "hex", "text_color": "hex"}
+    {"bold": true, "underlined": true, "italics": true, "font_size": 14, "background_color": <hex> color value, "text_color": <hex> color value}
```
Scott771, line 299: I think this is where Slater suggested "FFFFF" as the default sample, but to me this reads the same

dzereski: I talked this through with him and we agreed to just include <hex>

PR #21, opened 2020-04-07
Updated DocumentExtraction documentation for preset configurations

Scott771:
Added more information on what the preset configurations do and what python objects they return, removed a reference to the api token being recognized (without an explicit path) in the working directory, added some additional details/context for a couple of kwargs.


madisonmay:
I think in general we should bias towards code examples / truncated sample outputs when possible but the extra high level explanation certainly doesn't hurt either.

PR #22, opened 2020-04-08
catching most common object retrieval error 

Scott771:
When a user doesn't check if a Job.status is 'FAILURE' they are most likely going to run into a KeyError 'url'  when calling RetrieveStorageObject (most often when doing OCR w/ DocumentExtraction). Our Chatham clients (as well as myself and Andrew) have all run into this error and been initially stuck as to what the problem might be.

Updated with two lines to check if 'url' is in the storage object dictionary and, if not, raise an error for the user to check the job result/status and suggest that it could be a credential/specification issue. Even though this could be avoided in a user's own code, it has been a common problem already and seems that, for now, an easy fix with a better error message might save clients some frustration.

### DIFF
--- a/indico/queries/storage.py
+++ b/indico/queries/storage.py
@@ -20,6 +20,9 @@ class RetrieveStorageObject(HTTPRequest):
 
     def __init__(self, storage_object):
         if type(storage_object) == dict:
+            if "url" not in storage_object.keys():
+                raise KeyError("Unable to retrieve result. Please check the job status error for more details\
+                                and ensure that you are using the correct credentials and specifications for your use case")
             url = storage_object["url"]
         else:
             url = storage_object
--- a/indico/queries/storage.py
+++ b/indico/queries/storage.py
@@ -1,6 +1,7 @@
 import json
 from typing import List
 from indico.client.request import HTTPMethod, HTTPRequest
+from indico.errors import IndicoRequestError
 
 
 class RetrieveStorageObject(HTTPRequest):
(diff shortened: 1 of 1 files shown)


Scott771:
Current error message looks like: 
```
~/Desktop/jupyter_kernels/new_client/venv/lib/python3.7/site-packages/indico/queries/storage.py in __init__(self, storage_object)
     21     def __init__(self, storage_object):
     22         if type(storage_object) == dict:
---> 23             url = storage_object["url"]
     24         else:
     25             url = storage_object

KeyError: 'url'
```
In this particular case, the job failed due to a bad kwarg to docextract configuration.

indico/queries/storage.py
```diff
@@ -20,6 +20,9 @@ class RetrieveStorageObject(HTTPRequest):
 
     def __init__(self, storage_object):
         if type(storage_object) == dict:
+            if "url" not in storage_object.keys():
+                raise KeyError("Unable to retrieve result. Please check the job status error for more details\
```
jkgenser, line 24:
```
try:
    url = storage_object["url"]
except KeyError:
    raise IndicoRequestError()
```

Scott771: Ok- changed it to that format

indico/queries/storage.py
```diff
@@ -20,6 +20,9 @@ class RetrieveStorageObject(HTTPRequest):
...
+                                and ensure that you are using the correct credentials and specifications for your use case")
```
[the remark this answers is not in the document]
Scott771: Changed the message to include the above and removed mention of credentials

PR #30, opened 2020-04-15
FIX: base type list checks

madisonmay:


jkgenser:
Context here is that Madison was getting some decoding errors due to running on 3.6

madisonmay:
Maybe once we get this into CI we just run under 3.6 and 3.7

PR #31, opened 2020-04-16
ADD:predict options and test

jkgenser:
Update client test to call predict with threshold.


jkgenser:
Rebased against master branch so this is mergable. Also updated the predict confidence `threshold` to 0.25 to make sure we get predictions back when running the integration test.

PR #36, opened 2020-04-23
DEV-4416 - Python client doesn't batch dataset upload and hits HTTP status 413 (Request entity too large)

jkgenser:
Added batching support and client test. Load testing this functionality is an integration test. So long as size of image x batch size is under the request size limit, batching will result in not hitting the size limit.


jkgenser:
> I'm not sure we should make this image specific. I think we offer this for both image and pdf.

Updated to enable this.

indico/queries/datasets.py
```diff
@@ -186,12 +186,14 @@ def __init__(
         wait: bool = True,
         from_local_images: bool = False,
         image_filename_col: str = "filename",
+        image_upload_batch_size: int = 20,
```
nkassis, line 189: I think this should just be batch size ignore image

jkgenser: Updated. Latest commit also has some additional type annotation that was missing on `UploadBatched`

jkgenser: nm... I get that you meant not just changing name. It clicked that there are other data types that can be uploaded. Added this and test for pdf.

indico/types/base.py
```diff
@@ -17,26 +18,23 @@ def list_subtype(cls):
...
         return False
-    
+
     return (
-        (inspect.isclass(v) and issubclass(v, BaseType)) 
-        or v in [str, int, float, bool, JSONType] 
+        (inspect.isclass(v) and issubclass(v, BaseType))
+        or v in [str, int, float, bool, JSONType]
         or valid_type(list_subtype(v))
     )
```
jkgenser, line 27: This is all formatting.

PR #39, opened 2020-04-26
DEV-4178: Queries and tests for annotation and object detection metrics

jkgenser:
Also adds a client test for creating an annotation model


jkgenser:
Tests pass against dev.

indico/queries/metrics.py
```diff
@@ -0,0 +1,102 @@
+from indico.client.request import GraphQLRequest
```
nkassis, line 1: I think this probably should be a submodule of model_groups to make it clear what these are.

indico/queries/metrics.py
```diff
@@ -0,0 +1,102 @@
...
+from indico.types.metrics import SequenceMetrics
+import json
+
```
nkassis, line 4: We can probably add a GetModelMetrics query here that first gets the model group and type the gets the appropriate model group metric by type. Maybe make a hash of metrics by task_type so it's easy to resolve in the continuation.

indico/types/metrics.py
```diff
@@ -0,0 +1,70 @@
+from typing import List
```
nkassis, line 1: Similarly I think we may want to either submodule this to model_groups or add model_group in the name of the file for clarity

indico/queries/metrics.py
jkgenser: done.

indico/types/metrics.py
jkgenser: done.

indico/queries/metrics.py
jkgenser: Added the GetModelGroupMetrics query in the continuation style as prescribed

jkgenser:
Added retrain for metrics field. Client tests pass against dev after merging in refactored fog Evaluation schema.

indico/queries/model_groups/metrics.py
```diff
@@ -0,0 +1,135 @@
...
+        }
+      }
+    }
+    """
+
+    def __init__(self, model_group_id: int):
+        super().__init__(self.query, variables={"modelGroupId": model_group_id})
+
+    def process_response(self, response):
+        return json.loads(
```
nkassis, line 101: There no object for this one?

jkgenser: The object detection metrics are JSONString in the graphql schema

jkgenser:
Rebased

jkgenser:
> I think we should change the object detection metrics to a real schema (while keeping backwards for now) so we can make this interface nicer.

Will make JIRA for updating in graphql and then client.

PR #40, opened 2020-04-28
DEV-4139 not enough data cause model creation to hang

nkassis:


indico/queries/model_groups.py
```diff
@@ -233,7 +233,7 @@ def requests(self):
         if self.wait:
             req = GetModelGroupSelectedModelStatus(id=model_group_id)
             yield req
-            while self.previous not in ["FAILED", "COMPLETE"]:
+            while self.previous not in ["FAILED", "COMPLETE", "NOT_ENOUGH_DATA"]:
                 sleep(1)
```
jkgenser, line 237: you can use he `Debouncer` class here

jkgenser, line 237: Should we use debouncer here? It's not polling for job so it won't hammer api server, but is probably make a lot of REST calls

PR #42, opened 2020-05-04
DEV-4229 predict options optional and added test for images

nkassis:


indico/queries/model_groups.py
```diff
@@ -270,19 +272,39 @@ def process_response(self, response):
...
+    def _args_strings(self, **kwargs):
+        args = [k for k in self.query_args.keys() if kwargs.get(cc_to_snake(k))]
+
+        query_args_string = ",".join(f"${k}: {self.query_args[k]}" for k in args)
+        model_predict_args = ",".join(f"{k}: ${k}" for k in args)
+
+        query = self.query.replace("<QUERY_ARGS>", query_args_string)
+        query = query.replace("<MODEL_PREDICT_ARGS>", model_predict_args)
+
+        print(query)
```
jkgenser, line 297: remove print.

tests/integration/queries/test_document.py
```diff
@@ -116,3 +116,27 @@ def test_document_extraction_batched(indico):
...
+
+    extract = client.call(RetrieveStorageObject(job.result))
+
+    assert type(extract) == dict
+    assert "pages" in extract
+    image = extract["pages"][0]["image"]
+
+    image = client.call(RetrieveStorageObject(image))
+
+    assert image
```
madisonmay, line 142: Feel like this final assert should be an instance check or something

indico/queries/model_groups.py
madisonmay: rogue print

indico/queries/model_groups.py
```diff
@@ -270,19 +272,39 @@ def process_response(self, response):
...
         """
 
+    query_args = {
+        "modelId": "Int!",
+        "data": "[String]",
+        "predictOptions": "JSONString"
+    }
+
+    def _args_strings(self, **kwargs):
+        args = [k for k in self.query_args.keys() if kwargs.get(cc_to_snake(k))]
```
madisonmay, line 289: Generally think we should prefer loud errors to silent failures but this is internal so less critical

PR #45, opened 2020-05-27
Ben/add teach task creation

benleetownsend:


indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def __init__(self, id: int):
+        self.id = id
+        super().__init__()
+
+    def requests(self):
+        yield _GetDatasetInfo(self.id)
+
+
+class CreateDatasetTeach(RequestChain):
```
jkgenser, line 185: We should separate creating a dataset from creating a teach task. We should have `CreateTeachTask` that takes as argument an existing dataset and not always require the request chain to create a dataset first... or at least make it optional whether to create a dataset in this chain

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,227 @@
...
+    GraphQLRequest,
+    RequestChain,
+)
+
+from indico.types.questionnaire import Questionnaire, Example
+from indico.errors import IndicoNotFound
+
+
+class DataFileName(GraphQLRequest):
+    query = """
```
jkgenser, line 18: you should add docstrings to each of the new methods in the same format as our existing prebuilt queries. this will be automatically rendered by sphinx doc generator

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,244 @@
...
+            questionaire_id=questionaire_id, num_examples=self.num_examples
+        )
+        yield AddLabels(
+            target=[self.data.get(f.source) for f in self.previous],
+            dataset_id=self.dataset_id,
+            row_index=[d.row_index for d in self.previous],
+            labelset_id=labelset_id,
+        )
+
+        yield GetDataset(id=self.dataset_id)
```
jkgenser, line 244: Typically the final return value of a request that creates a resource is the id or object of the resource that was created

jkgenser:
This is looking good, adding Nic.

indico/queries/model_groups.py
```diff
@@ -110,14 +110,16 @@ class _CreateModelGroup(GraphQLRequest):
...
+            $modelTrainingOptions: JSONString,
+            $documentModel: Boolean,
         ) {
                 createModelGroup(
                     datasetId: $datasetId,
                     sourceColumnId: $sourceColumnId,
                     labelsetColumnId: $labelsetColumnId,
                     modelTrainingOptions: $modelTrainingOptions,
                     name: $name,
+                    documentModel: $documentModel,
```
nkassis, line 122: I would prefer this lives in the modelTrainingOptions, we don't have any other column here to pick a model type or alternatively be a much more generic model enum. Adding a boolean doesn't scale and this really is about picking the model_type.

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,244 @@
...
+
+class AddLabels(GraphQLRequest):
+    """
+    Add labels to an existing labelset.
+
+    Args:
+        dataset_id (int): The id of the dataset to add labels to.
+        labelset_id (int): The id of the labelset to add labels to.
+        target (List[Any]): A list of labels to add to the labelset.
+        row_index (List[Any]): The row indices corresponding with each item in targets.
```
nkassis, line 26: The interface here could require labels in the dictionary format. What made us choose the list version over just asking for the dictionary? I'm wondering if this interface might be more error prone and allow for mis ordered labels but was it more convenient in some contexts?

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,244 @@
...
+                ][0]["examples"]
+            ]
+        except IndexError:
+            raise IndicoNotFound(
+                "Examples not found. Please check the ID you are using."
+            )
+        return examples
+
+
+class CreateQuestionaire(GraphQLRequest):
```
nkassis, line 122: The creation is actually asynchronous and the questionnaire has a state. It's probably better to consider this a chain and have a wait keyword to check

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,244 @@
...
+        try:
+            questionnaire = Questionnaire(
+                **super().process_response(response)["createQuestionnaire"]
+            )
+        except IndexError:
+            raise IndicoNotFound("Failed to create Questionnaire")
+        return questionnaire
+
+
+class CreateTeachTask(RequestChain):
```
nkassis, line 184: The name here should not be CreateTeachTask it's not consistent with our naming convention. If the above create questionnaire is only for internal request  it should just be _CreateQuestionnaire and this one CreateQuestionnaire 

Also this query pushes into the realm that I'd rather leave as an example then a task. It's implying a lot about how it would be used in context (from a csv, labels would be available for examples etc...) that is more the realm of a script than the python client.

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,244 @@
...
+        yield CreateQuestionaire(
+            name=self.name,
+            dataset_id=self.dataset_id,
+            source_column_id=self.dataset.datacolumns[0].id,
+            targets=self.targets,
+        )
+        questionaire_id = self.previous.id
+        yield GetDataset(id=self.dataset_id)
+        labelset_id = self.previous.labelsets[0].id
+        yield GetQuestionaireExamples(
```
nkassis, line 234: We should make this optional if the labels are provided only.

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,244 @@
...
+        name: str,
+        csv_path: str,
+        dataset: Dataset,
+        num_examples: int,
+        text_column="text",
+        label_column="labels",
+    ):
+        self.dataset_id = dataset.id
+        self.dataset = dataset
+        csv = pd.read_csv(csv_path)
```
nkassis, line 213: I'm not a fan of the interface here, is there a reason we require a csv for this? It's limiting in context where the user doesn't have a file.

tests/integration/queries/test_teach_task.py
```diff
@@ -0,0 +1,33 @@
+import time
+import pytest
+from pathlib import Path
+from indico.client import IndicoClient
+from indico.queries.datasets import CreateDataset
+from indico.queries.teach_tasks import CreateTeachTask
+from indico.types.dataset import Dataset
+
+
+def test_create_dataset_teach(indico):
```
nkassis, line 10: Needs more tests

indico/queries/teach_tasks.py
```diff
@@ -0,0 +1,244 @@
...
+                id
+            }
+        }
+    """
+
+    def __init__(
+        self, name: str, dataset_id: int, source_column_id: int, targets: List[str]
+    ):
+        questions = [
+            {"type": "ANNOTATION", "targets": targets, "keywords": [], "text": name,}
```
nkassis, line 162: So we are only allowing annotation tasks? We can add an arg to the query for the type of question.

indico/queries/model_groups.py
jkgenser: @nkassis : modelTrainingOptions or a modelType enum per your comment on the fog PR?

indico/queries/questionnaire.py
```diff
@@ -0,0 +1,307 @@
...
+        Questionnaire object   
+    """
+
+    previous = None
+
+    def __init__(
+        self,
+        name: str,
+        targets: set,
+        dataset: Dataset,
```
nkassis, line 250: Can we use a union type of Dataset and Id here so we don't force the use of a dataset object and people can simply pass the id.

indico/queries/questionnaire.py
```diff
@@ -0,0 +1,307 @@
...
+
+    Raises:
+        IndicoInputError if the length of targets and row_index are not equal
+
+    """
+
+    query = """
+        mutation(
+            $labels: [SubmissionLabel]!,
+            $dataset_id: Int!,
```
nkassis, line 36: Unfortunate we need the dataset id here

PR #47, opened 2020-06-03
DEV-4283: Export query

jkgenser:


jkgenser:
Added Backend team to the PR so people can start getting familiar- but Nic has been the main code owner of the current python client.

indico/queries/export.py
```diff
@@ -0,0 +1,209 @@
...
+        super().__init__()
+
+    def requests(self):
+        yield _CreateExport(
+            dataset_id=self.dataset_id,
+            column_ids=self.column_ids,
+            subset_ids=self.subset_ids,
+            labelset_ids=self.labelset_ids,
+            file_info=self.file_info,
+            combine_labels=self.combine_labels,
```
pastap, line 201: I believe you're missing the anonymous kwarg here
and in the mutation above

jkgenser: will update.

indico/queries/export.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def __init__(self, export_id: int):
+        super().__init__(self.query, variables={"exportIds": [export_id]})
+
+    def process_response(self, response):
+        response = super().process_response(response)
+        return Export(**response["exports"]["exports"][0])
+
+
+class RetrieveExport(RetrieveStorageObject):
```
nkassis, line 112: Should this be _RetrieveExport? Do we expect people to call this directly?

indico/queries/export.py
```diff
@@ -0,0 +1,209 @@
...
+        self.export = export
+
+    def requests(self):
+        if self.export_id:
+            yield GetExport(self.export_id)
+        export = self.export or self.previous
+
+        if export.status != "COMPLETE":
+            raise IndicoRequestError(
+                code="JOB_INCOMPLETE",
```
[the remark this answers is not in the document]
jkgenser: updated

indico/queries/export.py
jkgenser: good point. updated to RetrieveExport. DownloadExport is the public query

PR #48, opened 2020-06-08
ADD: support for submission

hill-ba:


indico/queries/submission.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(
+        self, submission_id: int,
+    ):
+        super().__init__(
+            self.query, variables={"submissionId": submission_id},
+        )
+
+    def process_response(self, response) -> Job:
+        response = super().process_response(response)["submissionResults"]
+        return Job(id=response["jobId"])
```
pastap, line 68: do we need a request chain here?
like - createSubmissionResults followed by polling for the returned async job?

hill-ba: :man_shrugging: @astha seems like the polling for the job is part of the core client functionality. We could do that for users, but that seems pretty easy to do programatically as a user -- Thought I wouldn't say I have any context at all about what level of abstraction we're shooting for with the python client

pastap: oh cool! nah, you know more about the SDK than I do. this is solid then :+1: 
do you think we need some `SubmissonStatus` child class of JobStatus so we can poll for submission ids after a submission is created?

indico/types/submission.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Submission(BaseType):
+    """
+    A Submission in the Indico Platform
+
+    Attributes:
+        id (int): the Submission id
+        dataset_id (int): the Dataset id
+        workflow_id (int): the Workflow id
+        status (str): status of the submission
```
pastap, line 12: Going off of the pretty extensive docstring here: https://github.com/IndicoDataSolutions/indico-client-python/blob/master/indico/types/jobs.py
I think a lot more information can be added here?

indico/types/submission.py
```diff
@@ -0,0 +1,24 @@
...
+        input_file (str): local url to stored input
+        input_filename (str): name of the original file
+        result_file (str): local url to the stored output
```
pastap, line 15: Using previous comments as a guide, I think Indico URLs should be described as
> URL of the datafile within the Indico Platform

https://github.com/IndicoDataSolutions/indico-client-python/blob/master/indico/types/datafile.py#L14

indico/queries/submission.py
```diff
@@ -0,0 +1,68 @@
+import json
+from typing import Dict, List
+
+from indico.client.request import GraphQLRequest
+from indico.types import Job, Submission
+
+
+class ListSubmissions(GraphQLRequest):
```
pastap, line 8: I think this could really use a docstring - esp on how to use the filters.
I'd also make sure a default `limit` is supplied. iirc, if limit is not provided, filters are ignored. so we can have a default of 1000 or something

pastap:
Apologies, I know you kind of inherited my WIP commit on this branch for workflow submission creation. any chance you can zhuzh that up a bit with docstrings as well? it's #ugly

indico/types/submission.py
hill-ba: good comment, i had pulled these from Fog, but there's probably a more informative description to surface for SDK users.

indico/queries/submission.py
pastap: or maybe default limit if filters is supplied and limit isn't? or maybe we do that in Fog.... or maybe we fix atmos so folks can filter without limit :thinking: idk, what do you think?

indico/queries/submission.py
nkassis: Normally we add a wait flag to allow users to make it blocking.

indico/queries/submission.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+
+    def process_response(self, response) -> List[Submission]:
+        return [
+            Submission(**s)
+            for s in super().process_response(response)["submissions"]["submissions"]
+        ]
+
+
+class SubmissionResult(GraphQLRequest):
```
nkassis, line 49: This should be named CreateSubmissionResult to conform to naming convention.

indico/filters/base.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def __init__(self, filters: Dict[str, Any]):
+        self._validate(filters)
+        super().__init__(**filters)
+
+    @property
+    def filterable_columns(self):
+        return self._filterable_columns
+
+    @classmethod
```
pastap, line 26: why have these be class methods?

indico/filters/base.py
```diff
@@ -0,0 +1,34 @@
...
+    def _validate(cls, keys: Iterable):
+        if not all(key in cls._filterable_columns for key in keys):
+            raise TypeError()
+
+    @classmethod
+    def from_dict(cls, filters: Dict[str, Any]):
+        cls._validate(filters.keys())
+        return cls(filters)
```
pastap, line 34: this validates twice - once on the keys for filters and then again when you call `__init__` with `cls(filters)`. I wonder if `from_dict` is really a necessary function here?

indico/filters/base.py
```diff
@@ -0,0 +1,34 @@
...
+        self._validate(filters)
+        super().__init__(**filters)
+
+    @property
+    def filterable_columns(self):
+        return self._filterable_columns
+
+    @classmethod
+    def _validate(cls, keys: Iterable):
+        if not all(key in cls._filterable_columns for key in keys):
```
pastap, line 28: hmm, filters can take `AND` and `OR` as well
For example: https://github.com/IndicoDataSolutions/sunbow/blob/REPAIR/tests/routes/test_submission.py#L272
here I filter for 2 different statuses

indico/filters/submission.py
```diff
@@ -0,0 +1,5 @@
+from .base import BaseFilter
+
+
+class SubmissionFilter(BaseFilter):
```
pastap, line 4: if implemented class filters won't get more complicated than this, up to you but I think placing this class, and future BaseFilter subclasses into `indico/filters/__init__.py` could also work? this is totally up to you tho

indico/filters/base.py
```diff
@@ -0,0 +1,34 @@
...
+    Note:
+        Class should be extended for specific Indico resources, and the subclasses 
+        instantiated when a query is constructed.
+    """
+
+    _filterable_columns = set()
+
+    def __init__(self, filters: Dict[str, Any]):
+        self._validate(filters)
+        super().__init__(**filters)
```
pastap, line 20: `self.update(filters)`?

indico/queries/submission.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(
+        self,
+        submission_ids: List[int] = None,
+        workflow_ids: List[int] = None,
+        filters: Union[Dict, SubmissionFilter] = None,
+        limit: int = 1000,
+    ):
+        if isinstance(filters, dict):
+            filters = SubmissionFilter.from_dict(filters)
```
pastap, line 35: so this could just be `SubmissionFilter(filters)`

tests/integration/queries/test_submission.py
```diff
@@ -0,0 +1,26 @@
...
+    JobStatus,
+    ListSubmissions,
+    ListWorkflowsForDataset,
+    WorkflowSubmission,
+)
+from indico.types import Job, ModelGroup
+from ..data.datasets import airlines_dataset, airlines_model_group
+
+
+def test_list_submissions(indico, airlines_dataset, airlines_model_group: ModelGroup):
```
pastap, line 14: were you able to run this test against dev-repair?
credentials and API token pinned to the project-repair slack channel. you can also write tests for CreateSubmissionResult by going after any submission in the PENDING_REVIEW state - lmk and I can get you some ids you can test with:
workflow 1, submission 210
workflow 3, submission 205

indico/filters/base.py
hill-ba: this is just checking that the user is actually filtering on the correct columns, as defined by each particular resource.

tests/integration/queries/test_submission.py
hill-ba: I did test on repair for list submissions, ahven't yet tested for createSubmissionResult

indico/filters/base.py
hill-ba: I see what you mean ...

indico/filters/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+                f"Can only filter on {self._filterable_columns}, not {keys}"
+            )
+
+    def _update(self, key: str, value: Any, mode: FilterMode):
+        self._validate_keys([key])
+        existing_filters = self.pop(mode.name) if mode.name in self else None
+        filters = (
+            [*existing_filters, {key: value}] if existing_filters else [{key: value}]
+        )
+        self.update({mode.name: filters})
```
pastap, line 40: if you `return self` here, folks could do chains of like:
```
filter = SubmissionFilter({"status": "COMPLETE"}).and_("status", "COMPLETE")
```
which would be :fire:

indico/filters/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def and_(self, key, value):
+        return self._update(key, value, FilterMode.AND)
+
+    def or_(self, key, value):
+        return self._update(key, value, FilterMode.OR)
+
+    @property
+    def filterable_columns(self):
+        return self._filterable_columns
```
pastap, line 50: necessary?

tests/unit/test_filters.py
```diff
@@ -0,0 +1,30 @@
...
+
+    filter = SubmissionFilter({"status": "COMPLETE"})
+    with pytest.raises(ValueError):
+        filter.and_("name", "thing")
+
+
+if __name__ == "__main__":
+    test_submission_filter_and()
+    test_submission_filter_or()
+    test_invalid_filter_column()
```
pastap, line 30: :thinking: is this needed?

tests/integration/queries/test_submission.py
pastap: sounds like, based on the conversation today, integration tests for SubmissionResult will be in the integration-tests folder. so no need here

indico/queries/submission.py
```diff
@@ -0,0 +1,85 @@
...
+                }
+            }
+        }
+    """
+
+    def __init__(
+        self,
+        submission_ids: List[int] = None,
+        workflow_ids: List[int] = None,
+        filters: Union[Dict, SubmissionFilter] = None,
```
pastap, line 35: we could just accept a SubmissionFilter here, forcing the `filters` to go through validation, maybe?

hill-ba: Nic had suggested leaving it flexible for now -- I can request a final rewview from him after I implement your changes.

indico/filters/__init__.py
hill-ba: no but could be nice for folks toying in a interpreter and testing out things. I can remove.

indico/filters/__init__.py
hill-ba: good idea, done.

indico/queries/submission.py
pastap: nah, that's totally fair then. He's got final say :) I think in that case, maybe you'd want to do what you did before? where you created the SubmissionFilter object - to still be able to run those validation checks? totally up to you

hill-ba: that's a fair point

indico/filters/__init__.py
```diff
@@ -0,0 +1,51 @@
+from enum import Enum
+from typing import Any, Dict, Iterable
+
+
+class FilterMode(Enum):
+    AND = "and"
+    OR = "or"
+
+
+class BaseFilter(dict):
```
nkassis, line 10: Mmm so this a bit different then I expected in terms of interface, I was thinking the filters would be fields on a Filter object so like SubmissionFilter as a class with known filterable and safe fields.

I think the interface I was thinking is like

SubmissionFilter(inputFilename=["test.pdf"], status=["complete"])

or 

filter = SubmissionFilter()
filter.status = ["complete"] and if status has type annotation it would know how to figure it out.

indico/filters/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def and_(self, key, value):
+        return self._update(key, value, FilterMode.AND)
+
+    def or_(self, key, value):
+        return self._update(key, value, FilterMode.OR)
+
+
+class SubmissionFilter(BaseFilter):
+    _filterable_columns = {"inputFilename", "status"}
```
nkassis, line 51: I think these should be attributes on the class not just a set

indico/filters/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+    determined by the specific resource a user queries.
+
+    Note:
+        Class should be extended for specific Indico resources, and the subclasses 
+        instantiated when a query is constructed.
+    """
+
+    _filterable_columns = set()
+
+    def __init__(self, mapping: Dict[str, Any], mode: FilterMode = FilterMode.AND):
```
nkassis, line 23: I think instead of a Dict we use kwargs

indico/queries/submission.py
nkassis: My original idea here was to have known fields on a class like SubmissionFilter that would have type completion and the like and the dict alternative here was so if we had not yet added a field to the client it would be available to call (backup). The validation here prevents unknown fields which goes against my intent.

indico/types/submission.py
```diff
@@ -0,0 +1,30 @@
...
+        input_filename (str): name of the original file
+        result_file (str): URL of the result datafile within the Indico Platform
+ 
+    """
+
+    id: int
+    dataset_id: int
+    workflow_id: int
+    status: str
+    input_file: str
```
nkassis, line 28: What is this vs input_filename?

examples/submission.py
```diff
@@ -0,0 +1,26 @@
...
+config = IndicoConfig(''
+    host="app.indico.io", api_token_path="./path/to/indico_api_token.txt"
+)
+client = IndicoClient(config=config)
+
+# Create a new Submission using a previously created workflow
+workflow_id = 5
+job = client.call(
+    WorkflowSubmission(
+        workflow_id=workflow_id, files=["./path_to_doc.pdf"], submission=True
```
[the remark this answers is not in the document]
pastap: Absolutely :+1: 
honestly don't see a reason why anyone would use a non-submission flow from the SDK - could go so far as overriding this parameter to always be True - but setting it as default is a solid step

indico/types/submission.py
pastap: the rainbow URL

examples/submission.py
```diff
@@ -0,0 +1,26 @@
+from indico import IndicoClient, IndicoConfig
+from indico.queries import CreateSubmissionResult, WorkflowSubmission, JobStatus
+
+
+# Create an Indico API client
+config = IndicoConfig(''
+    host="app.indico.io", api_token_path="./path/to/indico_api_token.txt"
```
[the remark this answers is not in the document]
hill-ba: will change, but literally following suit from the other examples.

indico/queries/submission.py
hill-ba: Can you elaborate on this @nkassis , I'm not following 

```
 and the dict alternative here was so if we had not yet added a field to the client it would be available to call (backup)
```

hill-ba: - @nkassis, so you're suggesting we don't validate in the client, just let Fog do it?

indico/queries/submission.py
```diff
@@ -0,0 +1,85 @@
...
+        self.wait = wait
+        super().__init__(
+            self.query, variables={"submissionId": submission_id},
+        )
+
+    def process_response(self, response) -> Job:
+        response = super().process_response(response)["submissionResults"]
+        job = Job(id=response["jobId"])
+        if self.wait:
+            return JobStatus(id=job.id, wait=True).result
```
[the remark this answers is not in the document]
hill-ba: I added `wait` to the mutation because I believe you had suggested supporting that in the first round of reviews. I can modify this to follow existing patterns.

PR #49, opened 2020-06-09
Alias UploadImages #DEV-4624

madisonmay:


jkgenser:
This could work, but I'm not sure this is the best approach. Another idea is giving an optional argument to UploadDocument that defaults to False for backwards compat but if True, then appends URL_PREFIX? What do you think @nkassis

jkgenser:
that works too

madisonmay:
Can do -- I'll opt for the switcharoo.

PR #50, opened 2020-06-09
Madison/form preprocessing

madisonmay:


madisonmay:
Intentionally excluded from docs for now. I'll add a full .rst page for form extraction once this thing is tried and tested.

indico/queries/forms.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def process_response(self, response):
+        jobs = super().process_response(response)["activeFormFields"]["jobIds"]
+        if jobs:
+            return [Job(id=j) for j in jobs]
+        else:
+            return []
+
+
+# TODO: move into indico-client
```
nkassis, line 33: Isn't that done...?

indico/queries/forms.py
```diff
@@ -0,0 +1,60 @@
...
+    Attempt to auto-detect form fields and labels
+
+    Args:
+        files (str): list of filepaths to upload
+
+    Returns:
+        suggested_fields: list of dictionarys (1 per PDF) containing structured field data 
+    """
+
+    def __init__(
```
nkassis, line 45: Do we want to add a wait kwargs here like the other jobs?

madisonmay: Seems warranted.

indico/queries/forms.py
madisonmay: Hehehe -- yes.

indico/queries/forms.py
madisonmay: Ok on second glance it seems like none of the other queries that return jobs do this because JobStatus is the query with a wait kwarg.

madisonmay: If I added a wait kwarg would I still return jobs (just jobs that we already know are done / we have blocked on)?  Or have this be polymorphic and return job results when wait=True?

nkassis: OH right yeah right... hum I wrote it that way I should know ;p

PR #53, opened 2020-06-30
Madison/new form queries

madisonmay:


indico/queries/forms.py
```diff
@@ -58,3 +56,58 @@ def requests(self):
...
+    mutation listPrebuiltFormsMutation {
+        listPrebuiltForms {
+            forms {
+                id
+                name
+            }
+        }
+    }
+    """
+    method = HTTPMethod.GET
```
nkassis, line 73: This should be post all graphql requests are POST

indico/queries/forms.py
```diff
@@ -58,3 +56,58 @@ def requests(self):
         else:
             yield UploadDocument(files=self.files)
         yield _FormPreprocessing(files=self.previous)
+
+
+class ListPrebuiltForms(GraphQLRequest):
+
+    query = """
+    mutation listPrebuiltFormsMutation {
```
nkassis, line 64: by the name this should be a query not a mutation.

indico/queries/forms.py
```diff
@@ -58,3 +56,58 @@ def requests(self):
...
+        super().__init__(query=self.query)
+
+    def process_response(self, response):
+        return super().process_response(response)["listPrebuiltForms"]["forms"]
+
+
+class GetPrebuiltForm(GraphQLRequest):
+
+    query = """
+    mutation($form_id: Int!) {
```
nkassis, line 85: Same as above I think there some confusion on the way graphql works if these are built as mutations.

indico/queries/forms.py
```diff
@@ -58,3 +56,58 @@ def requests(self):
...
+                    bottom
+                    left
+                    right 
+                    pageNum
+                }
+            }
+        }
+    }
+    """
+    method = HTTPMethod.GET
```
nkassis, line 107: remove

indico/queries/forms.py
```diff
@@ -1,10 +1,11 @@
 from typing import List
 from pathlib import Path
 
-from indico.client.request import RequestChain, GraphQLRequest
+from indico.client.request import RequestChain, GraphQLRequest, HTTPMethod
```
jkgenser, line 4: unused import i believe

jkgenser:
We should add python types to convert the results of the new queries back into. And also add unit tests. See for example: https://github.com/IndicoDataSolutions/indico-client-python/blob/master/indico/queries/datasets.py#L53

PR #54, opened 2020-07-01
ADD: all submission capabilities

pastap:


examples/custom_client.py
```diff
@@ -0,0 +1,16 @@
...
+
+# Use environment variables
+os.environ["INDICO_PROTOCOL"] = "http"
+os.environ["INDICO_HOST"] = "foo.bar.com"
+
+# Use IndicoConfig
+my_config = IndicoConfig(
+    host="indico.my-company.com",  # Overrides environment variable
+    api_token_path="../path/to/custom_api_token.txt",
+    verify_ssl=False,
```
nkassis, line 12: that part shouldn't be necessary for most use cases we shouldn't show it ;p it's in the further docs

examples/get-training-model-progress.py
```diff
@@ -1,14 +1,10 @@
-from indico import IndicoClient, IndicoConfig
+from indico import IndicoClient
 from indico.queries import GetModelGroup, GetTrainingModelWithProgress
 
 # The model group ID can be found on the review page of the indico platform
 model_group_id = 4305
 
-my_config = IndicoConfig(
-    host="app.indico.io", api_token_path="./path/to/indico_api_token.txt"
```
nkassis, line 8: mmm I wonder if we remove this line, I think for like 90% of customer install they need to pass a config so worth showing?

indico/queries/workflow.py
```diff
@@ -1,49 +1,86 @@
...
+from indico.client.request import GraphQLRequest, RequestChain
 from indico.errors import IndicoError
 from indico.queries.storage import UploadDocument
-from indico.types.jobs import Job
-from indico.types.workflow import Workflow
+from indico.types import Job, Workflow
 
 
-class ListWorkflowsForDataset(GraphQLRequest):
+class ListAllWorkflows(GraphQLRequest):
```
nkassis, line 9: Lets name this to ListWorkflows. We can drop ListWorkflowsForDataset (I'll make this version 4.0 of the client so breaking change is ok+ I doubt anyone was using ListWorkflowForDataset)

examples/get-training-model-progress.py
pastap: I wanted to abstract all the redundant client config bits to the custom_client example

examples/custom_client.py
pastap: sure, will remove

examples/get-training-model-progress.py
nkassis: I'm questionning if that's a good idea for these examples, they can right now copy them as is, change the few fields that they need and go vs knowing how to get the config into the client. It's more verbose for sure.

pastap: I can put them back, no problem. But fwiw, I would argue that it's just as easy to look up the client example and copy that :woman_shrugging:

indico/queries/submission.py
```diff
@@ -0,0 +1,194 @@
...
+        mutation CreateSubmissionResults($submissionId: Int!) {
+            submissionResults(submissionId: $submissionId) {
+                jobId
+            }
+        }
+
+    """
+
+    def __init__(
+        self, submission_id: int
```
nkassis, line 123: you could also probably take a submission object here as an alternative to make it even easier. 

submission: Union[int, Submission] or just as two different params you check submission_id: int =None, submission: Submission

indico/queries/submission.py
```diff
@@ -0,0 +1,194 @@
...
+               `or result file Job did not complete within `timeout` seconds.
+        If not `wait`:
+            IndicoInputError: The requested Submission is not `check_status`
+    """
+
+    previous: Submission = None
+
+    def __init__(
+        self,
+        submission_id: int,
```
nkassis, line 162: same as above the interface could be a little nicer if we just took the submission object

indico/queries/submission.py
```diff
@@ -0,0 +1,194 @@
...
+        self.submission_id = submission_id
+        self.wait = wait
+        self.timeout = timeout
+        self.status_check = (
+            partial(eq, check_status) if check_status else partial(ne, "PROCESSING")
+        )
+
+    def requests(self) -> Union[Job, str]:
+        yield GetSubmission(self.submission_id)
+        debouncer = Debouncer(max_timeout=self.timeout)
```
nkassis, line 176: Won't this just increase until the next check is timeout time from the previous instead of timing out after timout time? This should just check like a set interval of time apart until done. The debouncer is more for preventing quick retries (albeit I'm not sure when it's really useful to backoff instead of just try every second outside like uploads)

examples/custom_client.py
```diff
@@ -0,0 +1,21 @@
+import os
+from indico import IndicoClient, IndicoConfig
+
+# Will connect to https://app.indico.io
+client = IndicoClient()
+
+# Environment variables override defaults
+os.environ["INDICO_PROTOCOL"] = "http"
```
nkassis, line 8: there no http support, https only

lizlemon: In the integration tests I create a client for http environments like this:

`IndicoClient(IndicoConfig(verify_ssl=False))`

lizlemon:
Should we be maintaining everything in tests/integration in this repository with each merge to master? This branch replaces `ListWorkfowsForDataset` with `ListWorkflows`, but tests/integration/queries/test_workflow.py still tries to import `ListWorkflowsForDataset`.

examples/custom_client.py
nkassis: Yeah config should not have that configurable, basically our authentication scheme only works with https in place.

nkassis:
> Should we be maintaining everything in tests/integration in this repository with each merge to master? This branch replaces `ListWorkfowsForDataset` with `ListWorkflows`, but tests/integration/queries/test_workflow.py still tries to import `ListWorkflowsForDataset`.

Yes @pastap that test needs updating. We now have the client in jenkins right @lizlemon so next step is adding those test runs to this repo so it blocks PRs correct?

examples/custom_client.py
pastap: except for single-node :D where we use http
but removed

examples/custom_client.py
```diff
@@ -0,0 +1,21 @@
+import os
+from indico import IndicoClient, IndicoConfig
+
+# Will connect to https://app.indico.io
+client = IndicoClient()
+
+# Environment variables override defaults
+os.environ["INDICO_PROTOCOL"] = "https"
```
nkassis, line 8: remove this from docs entirely

PR #55, opened 2020-07-06
explicitly mention OCR and fix location on how to pass docs in app

Scott771:
Small changes to workflow text in docs


dzereski:
Looks good

PR #56, opened 2020-07-07
fixing code snippet error

Scott771:
The code snippet in the docs for predictions included 'return' without any function. I also added in the imports as well as creating the client in case someone navigates to this page first


dzereski:
Looks fine

PR #60, opened 2020-07-13
added file type check to doc extraction

Scott771:
Currently, if you use DocumentExtraction with a file that isn't a PDF (the only currently supported doc type), e.g. .docx, you get a nondescript 500 error when you check the jobstatus. Changed the code to instead ensure that they are only uploading permissible documents initially. 

Current error at JobStatus
```
{'code': 500,
 'error_type': 'CyclibException',
 'message': 'http://apiserver:5000/async/a376ed6f-a00f-47ea-862a-3b696f896169: {}',
 'content': {},
 'error_class': 'ServiceException'}
```

### DIFF
--- a/indico/queries/documents.py
+++ b/indico/queries/documents.py
@@ -6,6 +6,7 @@ from typing import List
 from indico.client.request import RequestChain, GraphQLRequest, HTTPMethod, HTTPRequest
 from indico.types.jobs import Job
 from indico.queries.storage import UploadDocument, UploadBatched
+from indico.errors import IndicoError
 
 
 class _DocumentExtraction(GraphQLRequest):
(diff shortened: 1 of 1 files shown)


jkgenser:
Can you add a test that triggers this?

indico/queries/documents.py
```diff
@@ -80,6 +81,11 @@ def __init__(
         self.files = files
         self.json_config = json_config
         self.upload_batch_size = upload_batch_size
+        for f in self.files:
```
nkassis, line 84: I'm concerned with this check, this assumes all the pdf will have .pdf at the end which might not be the case in all customer envs. It's a very crude way to check.

Scott771: What do you think about something like below? The filetype.match(f) would potentially lead here to a FileNotFoundError- but I think that's probably better than also having that happen later as well. 
```
        for f in self.files:
            doctype = filetype.match(f)
            if doctype.MIME not in ('application/pdf',):
                raise IndicoError(
                    "Only PDF files can be used with DocumentExtraction. You provided: {f}"
                )
```

nkassis: yeah that should work, as long as it's consistent on windows and unix hopefully

dzereski: Shouldn't this work with TIF files as well?  Should also check ".tif", ".tiff"

jkgenser (commented):
stale?

PR #68, opened 2020-08-14
DEV-4870: updated dataset pipeline

jkgenser:
- CreateDataset request chain that creates dataset, adds, and processes files (maintains backwards compatibility with existing usages of CreateDataset)
- CreateEmptyDataset mutation that creates the empty dataset object for subsequent calls
- AddFiles for uploading and associate files with a dataset
- ProcessFiles - process datafiles request chain with wait option
- ProcessCSV - process csv request chain with wait option
- Tests that cover most use cases and basic errors to make sure errors are passed back to the client


indico/http/client.py
```diff
@@ -15,45 +14,47 @@
...
     def path_return_ok(self, path, request):
         return True
+
     def domain_return_ok(self, domain, request):
         return True
 
+
 class HTTPClient:
-    def __init__(self, config: IndicoConfig=None):
+    def __init__(self, config: IndicoConfig = None):
```
nkassis, line 33: Kind of weird it needs spaces when there is a type for not when it's just a kwarg.

indico/queries/datasets.py
```diff
@@ -306,3 +307,212 @@ def __init__(self, id):
...
     def process_response(self, response):
         return super().process_response(response)["deleteDataset"]["success"]
+
+
+#########################
+## Dataset Pipeline v2 ##
+#########################
+
+
+class CreateDataset_v2(GraphQLRequest):
```
nkassis, line 317: mmmm, I don't like this naming scheme... I think we should update the version of the client and make sure people know it only works with the new platform version.

indico/queries/datasets.py
```diff
@@ -306,3 +307,212 @@ def __init__(self, id):
...
+    query = """
+    mutation ($datasetId: Int!, $datafileIds: [Int]) {
+        addDataCsv(datasetId: $datasetId, datafileIds: $datafileIds) {
+            id
+            name
+        }
+    }
+    """
+
+    def __init__(self, dataset_id, datafile_ids):
```
nkassis, line 442: types

indico/queries/datasets.py
```diff
@@ -306,3 +307,212 @@ def __init__(self, id):
...
+    query = """
+    mutation($name: String!, $datasetType: DatasetType) {
+        createDataset(name: $name, datasetType: $datasetType) {
+            id
+            name
+        }
+    }  
+    """
+
+    def __init__(self, name, dataset_type=None):
```
nkassis, line 327: types

indico/queries/datasets.py
```diff
@@ -306,3 +307,212 @@ def __init__(self, id):
...
+        
+
+    Returns:
+        Dataset
+
+    Raises:
+
+    """
+
+    def __init__(
```
nkassis, line 468: types

indico/queries/datasets.py
```diff
@@ -306,3 +307,212 @@ def __init__(self, id):
...
+        wait (bool): Block while polling for status of files
+
+    Returns:
+        Dataset
+
+    Raises:
+
+    """
+
+    def __init__(self, dataset_id, datafile_ids, wait=True):
```
nkassis, line 503: types

indico/queries/datasets.py
jkgenser: OK. I'll have a `CreateDatasetPipeline` that will have similar behavior to current `CreateDataset` request chain for use in existing tests that use dataset as as fixture and for clients that want to do it in one operation.

indico/http/client.py
jkgenser: black formatter does the thinking so we don't have to debate it :shrug:  .... except line width

indico/queries/datasets.py
jkgenser: added

indico/queries/datasets.py
jkgenser: added

indico/queries/datasets.py
jkgenser: added

indico/queries/datasets.py
jkgenser: EDIT: after subsequent conversation, we discussed keeping `CreateDataset` with same API and additionally having `CreateEmptyDataset`

indico/queries/datasets.py
nkassis: Missing one for wait

jkgenser: added

PR #74, opened 2020-10-05
DEV-5300 Add ability to filter on retrieved flag

nkassis:
In addition I changed the default ordering for submissions to have most recent first. Retrieved is also now requested by default when listing submissions.


indico/filters/__init__.py
```diff
@@ -40,11 +40,16 @@ class SubmissionFilter(Filter):
     Returns:
         dict containing query filter parameters
     """
-    __options__ = ("input_filename", "status")
+    __options__ = ("input_filename", "status", "retrieved")
```
pastap, line 43: class docstring is missing information about this new thing

indico/filters/__init__.py
```diff
@@ -40,11 +40,16 @@ class SubmissionFilter(Filter):
...
-    def __init__(self, input_filename: str = None, status: str = None):
+    def __init__(self, input_filename: str = None, status: str = None, retrieved: bool = None):
         kwargs = {
             "input_filename": input_filename,
             "status": status.upper() if status else status,
         }
+
+
+        if retrieved != None:
+            kwargs["retrieved"] = retrieved
```
pastap, line 53: just add the key after line 48 - the Filter obj, in __init__, already does the "if not None" check

indico/queries/submission.py
```diff
@@ -66,6 +70,8 @@ def __init__(
                 "workflowIds": workflow_ids,
                 "filters": filters,
                 "limit": limit,
+                "orderBy": orderBy,
```
pastap, line 73: you actually need to modify the graphql filter to accept these vars

indico/queries/submission.py
```diff
@@ -13,14 +13,15 @@
...
-    List all Submissions visible to the authenticated user
+    List all Submissions visible to the authenticated user by most recent. 
 
     Args:
         submission_ids (List[int], optional): Submission ids to filter by
         workflow_ids (List[int], optional): Workflow ids to filter by
         filters (SubmissionFilter or Dict, optional): Submission attributes to filter by
         limit (int, optional): Maximum number of Submissions to return. Defaults to 1000
-
+        orderBy (str, optional): Order by field. Defaults to id
```
pastap, line 23: just to be pythonic - shouldn't this kwarg be snake case? the variable name passed to graphql can be camel case
see how submission_ids and workflow_ids are formatted

pastap (commented):
small docstring issue
other than that, lgtm

PR #82, opened 2020-10-20
NO MERGE INTENDED: model training options functional test example

jkgenser:
This test passes in scarlet. Customv2 has internal validation that will error if `n_epochs` is not between 1-16. Since I passed 0 in one case that model failed; this means that training options are being passed through to customv2 correctly.


jkgenser:
PR is for Liz to see an example of a functional test for this feature. Open debate on whether we want this test merged to master... implications of discussions on which tests go where.

PR #84, opened 2020-10-22
FIX: require some args to be keyword-only

pastap:
Basically don't allow folks to pass in args where we'd prefer kwargs (to prevent any unexpected results)
image

Also something worth adding is an "Options" part to the docstring to distinguish between required args and optional ones. could change that up in this PR - or we can decide on some standard and I can clean it all up in a separate PR

### DIFF
--- a/indico/queries/datasets.py
+++ b/indico/queries/datasets.py
@@ -2,29 +2,28 @@
 
 import json
 import tempfile
-import pandas as pd
 from pathlib import Path
-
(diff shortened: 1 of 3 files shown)


nkassis (approved):
@lizlemon some of your tests might break due to this, we made it so some optional args are no longer allowed just position ally, this is to avoid some sharp edges for devs where they accidentally pass ids in the wrong order.

PR #85, opened 2020-10-27
added export, filters,and submissions classes to documentation

Scott771:
added export, filters,and submissions classes to documentation using auto class/module


docs/_sources/review-submissions.rst.txt
```diff
@@ -0,0 +1,8 @@
+Review Submissions
```
pastap, line 1: question... why is this called "review submissions"? Submissions are just workflow submissions... not dependent on Review

docs/singleocr-ex.html
```diff
@@ -95,7 +95,10 @@
 <li class="toctree-l1"><a class="reference internal" href="model_predictions.html">Generating Model Predictions</a></li>
 <li class="toctree-l1"><a class="reference internal" href="workflow-intro.html">Workflows</a></li>
 <li class="toctree-l1"><a class="reference internal" href="review-intro.html">Review</a></li>
+<<<<<<< HEAD
+=======
```
pastap, line 99: probably not intended :)

docs/_sources/review-submissions.rst.txt
Scott771: My initial feeling was that while submissions dont require review, most of the functionality of that module is around submissions to review and that new users would find "Submissions" too ambiguous. Don't feel strongly though, so changed it

docs/singleocr-ex.html
Scott771: removed

docs/_sources/review-submissions.rst.txt
pastap: gotcha - I see what you mean. I think either submissions or workflow-submissions would totally work.
side note - should this file get deleted since you have a new `submissions.rst` file?

docs/client.html
```diff
@@ -106,6 +106,7 @@
 <li class="toctree-l1"><a class="reference internal" href="datasets.html">Datasets</a></li>
 <li class="toctree-l1"><a class="reference internal" href="model_groups.html">Model Groups</a></li>
 <li class="toctree-l1"><a class="reference internal" href="jobs.html">Jobs and Job Status</a></li>
+<li class="toctree-l1"><a class="reference internal" href="review-submissions.html">Review Submissions</a></li>
```
pastap, line 109: do we want to switch all of these as well?

docs/singleocr-ex.html
```diff
@@ -117,8 +117,13 @@
 <li class="toctree-l1"><a class="reference internal" href="datasets.html">Datasets</a></li>
 <li class="toctree-l1"><a class="reference internal" href="model_groups.html">Model Groups</a></li>
 <li class="toctree-l1"><a class="reference internal" href="jobs.html">Jobs and Job Status</a></li>
+<li class="toctree-l1"><a class="reference internal" href="review-submissions.html">Review Submissions</a></li>
 <li class="toctree-l1"><a class="reference internal" href="storage.html">Storage</a></li>
 <li class="toctree-l1"><a class="reference internal" href="workflow.html">Workflows</a></li>
+<<<<<<< HEAD
```
pastap, line 123: this is in a couple places

docs/storage.html
```diff
@@ -250,7 +262,7 @@
         <a href="workflow.html" class="btn btn-neutral float-right" title="Workflows" accesskey="n" rel="next">Next <span class="fa fa-arrow-circle-right"></span></a>
       
       
-        <a href="jobs.html" class="btn btn-neutral float-left" title="Jobs and Job Status" accesskey="p" rel="prev"><span class="fa fa-arrow-circle-left"></span> Previous</a>
+        <a href="review-submissions.html" class="btn btn-neutral float-left" title="Review Submissions" accesskey="p" rel="prev"><span class="fa fa-arrow-circle-left"></span> Previous</a>
```
pastap, line 265: like, in this context, I think Workflow Submissions... you're right  - would be less ambiguous than just `Submissions`

docs/client.html
Scott771: Hmmm- not sure why Sphinx didn't automatically update all of the references and delete that html all together, will investigate

docs/_sources/review-submissions.rst.txt
Scott771: sphinx-build doesn't delete old html files automatically- went and manually removed these

docs/client.html
Scott771: these have all been updated

docs/singleocr-ex.html
Scott771: these are gone

docs/storage.html
Scott771: I think it's ok to stick with just submissions for now- the application of these classes will be demonstrated in the initial docs under Workflows / Review etc.

dzereski (approved):

PR #86, opened 2020-10-30
[DEV-5350] ADD: submission auto review queries

pastap:
```
Waited submissions [(11441, 'PENDING_AUTO_REVIEW', '/storage/submission/1632/11441/submission_11441_result.json'), (11442, 'COMPLETE', '/storage/submission/1632/11442/submission_11442_result.json')]
Changes {'invoices price range q450 model': {'confidence': {'more_than_1000': 0.7106317558145944, '500_to_1000': 0.2126695627565185, 'less_than_500': 0.07881872964357038}, 'label': ['more_than_1000'], 'accepted': True}, 'invoices_extraction_less_than_500': None, 'invoices_extraction_more_than_1000': [{'start': 0, 'end': 18, 'label': 'Vendor', 'text': 'InsureTech Connect', 'confidence': {'<PAD>': 0.000531987810973078, 'Vendor': 0.9990828037261963, 'Line Item': 0.00023377692559733987, 'Line Item Value': 9.400565613759682e-05, 'Total': 1.2766715371981263e-05, 'Invoice Number': 4.4721000449499115e-05}, 'accepted': True}, {'start': 173, 'end': 176, 'label': 'Invoice Number', 'text': '122', 'confidence': {'<PAD>': 0.0006733918562531471, 'Vendor': 5.276270258036675e-06, 'Line Item': 1.663380317040719e-05, 'Line Item Value': 8.566633368900511e-06, 'Total': 2.9603956136270426e-05, 'Invoice Number': 0.9992665648460388}, 'accepted': True}, {'start': 386, 'end': 433, 'label': 'Line Item', 'text': 'InsureTech Connect 2018 Sponsor/Exhibit Package', 'confidence': {'<PAD>': 0.001972711645066738, 'Vendor': 0.00023570473422296345, 'Line Item': 0.9976629018783569, 'Line Item Value': 7.42916963645257e-05, 'Total': 3.6635203287005424e-05, 'Invoice Number': 1.769045411492698e-05}, 'accepted': True}, {'start': 525, 'end': 535, 'label': 'Line Item Value', 'text': '$ 7,591.00', 'confidence': {'<PAD>': 0.001199838938191533, 'Vendor': 1.1854126569232903e-05, 'Line Item': 1.712173252599314e-05, 'Line Item Value': 0.9982567429542542, 'Total': 0.0004936189507134259, 'Invoice Number': 2.078176839859225e-05}, 'accepted': True}, {'start': 697, 'end': 708, 'label': 'Total', 'text': '$ 10,845.00', 'confidence': {'<PAD>': 0.0002703087229747325, 'Vendor': 7.571233436465263e-05, 'Line Item': 0.00020587278413586318, 'Line Item Value': 0.00042587026837281883, 'Total': 0.9987584352493286, 'Invoice Number': 0.00026386132230982184}, 'accepted': True}]}
b362bdd0-b1f8-4e1d-8ba8-95c845639a81 has result {'review_id': 768, 'success': True}
```

Looks good - and comes up from graphql correctly
image

### DIFF
--- a/examples/submission.py
+++ b/examples/submission.py
@@ -6,7 +6,9 @@ from indico.queries import (
     ListSubmissions,
     RetrieveStorageObject,
     SubmissionResult,
+    SubmitReview,
     UpdateSubmission,
+    WaitForSubmissions,
     WorkflowSubmission,
     WorkflowSubmissionDetailed,
 )
@@ -75,3 +77,33 @@ result = client.call(RetrieveStorageObject(result_url.result))
 print(result)
 
 client.call(UpdateSubmission(submission.id, retrieved=True))
+
+
+"""
+Example 4
+Submit "auto-review" for a submission
+Requires auto-review is enabled for workflow
+"""
+submission_ids = client.call(
+    WorkflowSubmission(
+        workflow_id=workflow_id, files=["my_file.pdf"]
+    )
+)
+submissions = client.call(WaitForSubmissions(submission_ids))
+submission = submissions[0]
+raw_result = client.call(RetrieveStorageObject(submission.result_file))
+raw_result = client.call(RetrieveStorageObject(submission.result_file))
+changes = raw_result["results"]["document"]["results"]
+rejected = False
+for model, preds in changes.items():
+    if isinstance(preds, dict):
+        preds["accepted"] = True
+    elif isinstance(preds, list):
+        for pred in preds:
+            pred["accepted"] = True
+    else:
+        rejected = True
+print(f"Submitting review for {submission.id}: {changes}")
+job = client.call(SubmitReview(submission.id, changes=changes, rejected=rejected))
+job = client.call(JobStatus(job.id))
+print("Review", job.id, "has result", job.result)
--- a/indico/queries/submission.py
+++ b/indico/queries/submission.py
@@ -1,3 +1,4 @@
+import json
 import time
 from functools import partial
 from operator import eq, ne
@@ -65,7 +66,7 @@ class ListSubmissions(GraphQLRequest):
         filters: Union[Dict, SubmissionFilter] = None,
         limit: int = 1000,
         order_by: str = "ID",
-        desc: bool = True
+        desc: bool = True,
     ):
         super().__init__(
             self.query,
@@ -75,7 +76,7 @@ class ListSubmissions(GraphQLRequest):
                 "filters": filters,
                 "limit": limit,
                 "orderBy": order_by,
-                "desc": desc
+                "desc": desc,
             },
         )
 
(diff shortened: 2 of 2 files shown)


indico/queries/submission.py
```diff
@@ -124,6 +125,58 @@ def process_response(self, response) -> Submission:
...
+        self, submission_ids: List[int], timeout: Union[int, float] = 60,
+    ):
+        if not submission_ids:
+            raise IndicoInputError("Please provide submission ids")
+
+        self.submission_ids = submission_ids
+        self.timeout = timeout
+        self.status_check = partial(ne, "PROCESSING")
+        self.status_getter = partial(ListSubmissions, self.submission_ids, limit=None)
+
```
sihrc, line 165: We wouldn't happen to have query for just status would we?

pastap: nope - just GetSubmission for the whole object and ListSubmissions for multiple

examples/submission.py
```diff
@@ -75,3 +77,33 @@
...
+        workflow_id=workflow_id, files=["my_file.pdf"]
+    )
+)
+submissions = client.call(WaitForSubmissions(submission_ids))
+submission = submissions[0]
+raw_result = client.call(RetrieveStorageObject(submission.result_file))
+raw_result = client.call(RetrieveStorageObject(submission.result_file))
+changes = raw_result["results"]["document"]["results"]
+rejected = False
+for model, preds in changes.items():
```
nkassis, line 98: This could use some comments since I presume it's based on the model type?

indico/queries/submission.py
nkassis: We could make a hidden query here to reduced the fields requested (since list gets a lot more). This would only get you up to Fog where Submission status resolver would have to be written with this in mind. We could make the first portion of making the client ask for less and the second portion if we detect an issue server side with these queries. 

Additionally I think eventually we need to start having get multiple queries that don't use the list method. Andy asked for this as he pointed out they get really annoying with the paging if you aren't doing paging and just getting items by ids. But that's a tangent I think for this one we just start with a GetSubmissionStatuses query that can take a list of ids and return objects with just the status, id, and errors filed in.

indico/queries/submission.py
```diff
@@ -262,3 +316,53 @@ def requests(self) -> Union[Job, str]:
...
+            variables={
+                "submissionId": submission_id,
+                "changes": changes,
+                "rejected": rejected,
+            },
+        )
+
+    def process_response(self, response) -> Job:
+        response = super().process_response(response)["submitAutoReview"]
+        return Job(id=response["jobId"])
```
nkassis, line 368: Is the posting of reviews async? Is that the merge operation server side?

pastap: it's async - know it's been a while but we talked about this - needs to be async bc we need to validate the review before saving

nkassis: Ah ok makes sense.

indico/queries/submission.py
pastap: hmmm, for Andy... you don't have to paginate - could just not provide a `limit`
maybe it's part of a bigger ask to filter without limits? not sure... but ListSubmissions does work without pagination

As for here - usually when we have these "limited field" queries, we need a server-side query to support it first (since the graphql query name is different) before we use it in the client. I can make the server-side query if we'd like - my thinking was that a submission "row" is small enough as is that querying for the whole thing isn't too bad, but can see how that might cause huge issues later with the high load we're seeing.

nkassis: It's more that the nested object is a pain to deal with particularly how it affects their caching layer. He can explain if you ask him.

nkassis (approved):

PR #88, opened 2020-12-01
[DEV-5634] ADD: force_complete kwarg to SubmitReview

pastap:
Depends on:
- https://github.com/IndicoDataSolutions/fog/pull/263

To be releases in 4.6


indico/queries/submission.py
```diff
@@ -328,16 +328,31 @@ class SubmitReview(GraphQLRequest):
...
+            $submissionId: Int!,
+            $changes: JSONString,
+            $rejected: Boolean,
+            $forceComplete: Boolean
+        ) {
+            submitAutoReview(
+                submissionId: $submissionId,
+                changes: $changes,
+                rejected: $rejected,
+                forceComplete: $forceComplete
```
nkassis, line 354: we need to move some of these to add them to the query if set otherwise we needlessly break backwards compatibility. I presume the default is false and isn't a required arg?

pastap: yes, not a required arg - so shouldn't break backwards compat...

pastap: oh but with fog - hmmm, I see... yeah, not sure - I think this was part of the convo around platform/client versioning and release notes. what do you think here? try with the kwarg and handle "invalid query" errors by removing some kwargs?

jkgenser (approved):

indico/queries/submission.py
pastap: oh cool! alright - will switch to using this :+1:

nkassis (approved):

PR #89, opened 2020-12-02
catch index error from bad id

Scott771:
Currently if you pass an invalid model group ID to ObjectDetectionMetrics, you get an IndexError (result returns empty from grapqhl query). Catching that error and raising a helpful exception message.

### DIFF
--- a/indico/queries/model_groups/metrics.py
+++ b/indico/queries/model_groups/metrics.py
@@ -98,11 +98,20 @@ class ObjectDetectionMetrics(GraphQLRequest):
         super().__init__(self.query, variables={"modelGroupId": model_group_id})
 
     def process_response(self, response):
-        return json.loads(
-            super().process_response(response)["modelGroups"]["modelGroups"][0][
-                "selectedModel"
-            ]["evaluation"]["metrics"]
-        )
+        try:
+            return json.loads(
+                super().process_response(response)["modelGroups"]["modelGroups"][0][
+                    "selectedModel"
+                ]["evaluation"]["metrics"]
+            )
+        except IndexError:
+            raise IndicoRequestError(
+                (
+                    "No results found. Please check that you have used the correct Model Group ID "
+                    "(*not* the selected model ID) and have permission to access that model."
+                ),
+                code=400,
+            )
 
 
 task_type_query_mapping = {
--- a/tests/integration/queries/test_model_group.py
+++ b/tests/integration/queries/test_model_group.py
@@ -226,7 +226,7 @@ def test_object_detection_metrics_bad_id(indico):
     from indico import IndicoConfig
     client = IndicoClient()
     with pytest.raises(IndicoRequestError):
-        _ = client.call(ObjectDetectionMetrics(-1))
+        client.call(ObjectDetectionMetrics(-1))
 
 
 def test_model_group_metrics_query(
--- a/tests/integration/queries/test_model_group.py
+++ b/tests/integration/queries/test_model_group.py
@@ -30,7 +30,7 @@ from ..data.datasets import (
     org_annotate_model_group,
     org_annotate_dataset,
 )
-from indico.errors import IndicoNotFound
+from indico.errors import IndicoNotFound, IndicoRequestError
 
 
 def test_create_model_group(airlines_dataset: Dataset):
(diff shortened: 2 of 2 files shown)


tests/integration/queries/test_model_group.py
```diff
@@ -222,6 +222,13 @@ def test_object_detection_metrics(
         assert isinstance(result["bbox"][metric_type], float)
 
 
+def test_object_detection_metrics_bad_id(indico):
+    from indico import IndicoConfig
+    client = IndicoClient()
+    with pytest.raises(IndicoRequestError):
+        _ = client.call(ObjectDetectionMetrics(-1))
```
jkgenser, line 229: why `_ =` instead of just `client.call(...)`

jkgenser: you can do `pytest.raises(Error) as e` and then make assertions on `e` to make sure it has the desired error message

Scott771: mostly because stylistically it is more akin to how that line of code would be called (i.e. returning the object into a variable), obviously (should) never gets there

pastap:
hmm... might be worth just catching and raising this error on the backend side, no? like if modelgroupid doesn't exist, raise 404

indico/queries/model_groups/metrics.py
```diff
@@ -98,11 +98,20 @@ def __init__(self, model_group_id: int):
         super().__init__(self.query, variables={"modelGroupId": model_group_id})
 
     def process_response(self, response):
-        return json.loads(
-            super().process_response(response)["modelGroups"]["modelGroups"][0][
-                "selectedModel"
-            ]["evaluation"]["metrics"]
-        )
+        try:
```
jkgenser, line 101: If there are no responses returned, graphql should return `[]` or something similar. I think a cleaner way would be to do that and then if no model group is found matching the id, then raise the error instead of catching index error. Index error is on `model_groups[0]` but it might be obscured by catching any index error in the entire block

Scott771: Object returns like: 

`{'data': {'modelGroups': {'modelGroups': []}}}`

Scott771: to me checking that [] empty vs. indexerror (which would always be that it is empty), same... but happy to change if you prefer that

Scott771 (commented):
> hmm... might be worth just catching and raising this error on the backend side, no? like if modelgroupid doesn't exist, raise 404
> like it might be worth adding a query "getModelGroup" that will raise the error
> also tho... this might be a bug if we don't raise an error for missing model groups

Yup- could also be an permission issue in addition to non-existent

pastap:
> > hmm... might be worth just catching and raising this error on the backend side, no? like if modelgroupid doesn't exist, raise 404
> > like it might be worth adding a query "getModelGroup" that will raise the error
> > also tho... this might be a bug if we don't raise an error for missing model groups
> 
> Yup- could also be an permission issue in addition to non-existent

Possibly - imo, I think this shouldn't be handled at all in the client - if you have a bug ticket, lmk, otherwise I can create one to make sure we raise errors on invalid model group ids

Scott771:
@pastap created here: https://indicodata.atlassian.net/browse/DEV-5636
I think it most common error for people to mess up is Model Group ID and Selected Model ID- SDK user wont get a hint that is why they're getting 404- but I defer

pastap:
> @pastap created here: https://indicodata.atlassian.net/browse/DEV-5636
> I think it most common error for people to mess up is Model Group ID and Selected Model ID- SDK user wont get a hint that is why they're getting 404- but I defer

Sweet, thank you!
I think that's great feedback - and of course... it's always going to be tricky if a model group and model exist with the same id - then the user might think they're doing the right thing bc no error... until later down the line. If you have any suggestions or ideas for how to combat that, I'd be down to work on a solution there!

indico/queries/model_groups/metrics.py
nkassis: so that endpoint filters instead of returning an error if not found same way it returns nothing if filtering by dataset id and no model groups are found. I think the logic below is fine or could be done with out try except by checking len and that the model group id matches. Instead of trusting the indexing error.

Long term we will make a direct way to get model groups that would offer errors for missing items but I think this patch is functional for now.

PR #90, opened 2020-12-03
[DEV-5635] ADD: pagination to client

pastap:
Basically creates a generator for users to loop through pages for List<> queries. Implemented on ListSubmissions since that was the ask from clients (and I don't think any other query will exceed 1000 items, tbh)
Atm, only supports going forwards - don't see a need yet to support stuff like "before" and so on

### DIFF
--- a/indico/client/client.py
+++ b/indico/client/client.py
@@ -5,7 +5,7 @@ import urllib3
 
 from indico.config import IndicoConfig
 from indico.http.client import HTTPClient
-from indico.client.request import HTTPRequest, RequestChain
+from indico.client.request import HTTPRequest, RequestChain, PagedRequest
 
 
 class IndicoClient:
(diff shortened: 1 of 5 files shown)


tests/integration/queries/test_workflow.py
```diff
@@ -138,6 +138,24 @@ def test_list_workflow_submission_retrieved(
...
+    client = IndicoClient()
+    wfs = client.call(ListWorkflows(dataset_ids=[airlines_dataset.id]))
+    wf = max(wfs, key=lambda w: w.id)
+
+    dataset_filepath = str(Path(__file__).parents[1]) + "/data/mock.pdf"
+
+    submission_ids = client.call(
+        WorkflowSubmission(workflow_id=wf.id, files=[dataset_filepath]*5)
+    )
+    for sub in client.paginate(ListSubmissions(workflow_ids=[wf.id], limit=3)):
```
nkassis, line 153: If we pass a limit of 3 we should get 3 back, if they don't pass this limit we should pass one or use the server default.

indico/client/client.py
```diff
@@ -64,3 +64,16 @@ def call(self, request: Union[HTTPRequest, RequestChain]):
             return self._handle_request_chain(request)
         elif request and isinstance(request, HTTPRequest):
             return self._http.execute_request(request)
+
+    def paginate(self, request: PagedRequest):
```
nkassis, line 68: I don't think we need this function, PageRequest can encapsulate this.

PR #94, opened 2021-01-19
[DEV-5645] ADD: AddDataToWorkflow SDK Query

thearchitector:


pastap (approved):
lgtm - adding Jerry and Nic

tests/integration/queries/test_workflow.py
```diff
@@ -210,3 +215,58 @@ def test_workflow_submission_auto_review(
     job = client.call(JobStatus(job.id))
     submission = client.call(GetSubmission(sub.id))
     assert submission.status == "COMPLETE" if force_complete else "PENDING_REVIEW"
+
+
+def _new_dataset_for_updating(client):
+    # new dataset
+    airline_csv = str(Path(__file__).parents[1]) + "/data/AirlineComplaints.csv"
```
pastap, line 222: do any preexisting tests fail because of the csv column reordering?

thearchitector: none that i can tell

indico/queries/workflow.py
```diff
@@ -280,3 +281,59 @@ def __init__(
         self, workflow_id: int, files: List[str] = None, urls: List[str] = None
     ):
         super().__init__(workflow_id, files=files, urls=urls, submission=True)
+
+
+class _AddDataToWorkflow(GraphQLRequest):
+    query = """
+        mutation addDataToWorkflow($workflowId: Int!) {
+            addDataToWorkflow(workflowId: $workflowId){
+                subsetId
```
nkassis, line 290: Isn't subset id set async? What's the point of including it here?

tests/integration/queries/test_workflow.py
sihrc: and here I've been using ".." / ".."

indico/queries/workflow.py
```diff
@@ -280,3 +281,59 @@ def __init__(
...
+
+    Options:
+        wait (bool, default=False): Block while polling for status of update
+
+    Returns:
+        Workflow: Updated Workflow object
+    """
+
+    def __init__(self, workflow: int, wait: bool = False):
+        self.workflow_id = workflow.id if isinstance(workflow, Workflow) else workflow
```
sihrc, line 328: I'm kind of sad this is how we've treated workflow_id. 
I think ideally for me , we'd have self.workflow and a self.workflow_id instead of having workflow_id potentially be a whole ass workflow

sihrc (approved):

jkgenser (approved):

indico/queries/workflow.py
thearchitector: yea, tbh i didn't think it made much sense but just followed what was there. might do what you say anyways, though. in general, i think that the queries need to stick to either id or the object, just cause having both is pretty confusing from a usability perspective (totally unclear why I would be able to use a workflow in one query but not another), etc

thearchitector: fundamentally i think it comes down to whether we want customers to think in terms of engineering or "Indico as a product"

from an engineering pov, makes way more sense to just use the id. from a ux one, id say the whole object (even if the guts just use the id), just because the actions are being done on Workflows, not workflow ids

indico/queries/workflow.py
pastap: @nkassis so the subset_id that comes back from AddDataToWorkflow is the new subset_id that the workflow is going to be working with - do you think it's worth returning/exposing?

thearchitector: Functionally, what can you do with it? If it's just extraneous then returning it wouldn't be super useful besides data management on the client side (though that might be a reason in and of itself)

pastap: no real functionality other than information - ok without it

nkassis (approved):

PR #97, opened 2021-02-17
ADD: Support for multi-file submissions and result versions

thearchitector:


indico/queries/workflow.py
```diff
@@ -134,23 +133,27 @@ def requests(self):
 class _WorkflowSubmission(GraphQLRequest):
 
     query = """
-        mutation workflowSubmissionMutation($workflowId: Int!, ${arg_}: {type_}, $recordSubmission: Boolean) {{
-            {mutation_name}(workflowId: $workflowId, {arg_}: ${arg_}, recordSubmission: $recordSubmission) {{
+        mutation workflowSubmissionMutation($workflowId: Int!, ${arg_}: {type_}, $recordSubmission: Boolean, $bundle: Boolean, $resultVersion: SubmissionResultVersion) {{
+            {mutation_name}(workflowId: $workflowId, {arg_}: ${arg_}, recordSubmission: $recordSubmission, bundle: $bundle, resultVersion: $resultVersion) {{
```
jkgenser, line 137: If someone on an older IPA version runs this query and `resultVersion` is not on the schema this will break. How are you guys thinking of mitigating this?

pastap: This is a super good point. Elias, here is an example of how we dealt with exposing an option to clients without expecting that the underlying server might not support it (in this case, predictOptions was the issue): https://github.com/IndicoDataSolutions/indico-client-python/blob/master/indico/queries/model_groups/model_groups.py#L291
You can see, Jerry had to basically format the graphene query string based upon user input - so "bundles" as an arg only gets used if "bundles" kwarg is specified by a user

thearchitector: This is only a problem because customers might be running their own platform right? This wouldn't be a problem we were in control

jkgenser: Even if we run the platform, the customer may be on an older version because the customer or us hasn't schedule an upgrade

thearchitector: What would a case where we require a the clients be used with certain ipa version ranges? Having the clients support all ipa versions doesn't make much sense to me

jkgenser: We don't right now to my knowledge have explicit pinning of what IPA platform versions a given client version supports or how many versions back a new client version should support but these are good points. I know it's been discussed but I don't think a final solution has been settled on. IMO it's hard to enforce this until we have automated testing set up for clients which we don't yet. And also platforms versions for them to run against. But Liz and Nic would better in terms of what policies we'll drive for and what constraints are preventing us from having created or enforcing them yet

jkgenser (approved):

pastap (approved):

indico/queries/workflow.py
```diff
@@ -278,9 +322,21 @@ class WorkflowSubmissionDetailed(WorkflowSubmission):
     detailed_response = True
 
     def __init__(
-        self, workflow_id: int, files: List[str] = None, urls: List[str] = None
+        self,
```
nkassis, line 325: Did the black formatter let this happen? I like this style but the formatter used to just change this to all be on one line.

tests/integration/queries/test_workflow.py
```diff
@@ -90,12 +90,77 @@ def test_workflow_submission(
...
+        {"urls": [PUBLIC_URL + "mock.pdf"] * 3},
+        {"files": [str(Path(__file__).parents[1]) + "/data/mock.pdf"] * 3},
+    ],
+)
+def test_workflow_submission_versioned(
+    indico, airlines_dataset, airlines_model_group: ModelGroup, _input
+):
+    client = IndicoClient()
+    wfs = client.call(ListWorkflows(dataset_ids=[airlines_dataset.id]))
+    wf = max(wfs, key=lambda w: w.id)
```
nkassis, line 112: kinda feels like we should put this into a fixture, also probably not trust whatever the latest workflow is on that dataset. Could be something crazy and we now have a test interdependence.

nkassis (approved):
Some minor comments that aren't blocking but I think we could do them separately as improvements.

indico/queries/workflow.py
thearchitector: yea, it formatted it - not sure why but thats what it decided was best

PR #98, opened 2021-02-25
Codeship CI/CD for python client

goatrocks:
This uses enables CI/CD for the python client using versioneer and codeship. 

~~Currently this draft PR pushes to test.pypi.org, with two test pushes here done using jet locally: https://test.pypi.org/project/indico-client/#history~~

Updated to have creds for pypi.org.

Here's a shot of the tests running in codeship:
image

The tests will run on any push right now, but can be configured to just run on PR.

The deploy step will only run on creation of a new tag following the x.y.z versioning. Pypi expects pep440 version strings, so pre-release clients can be released by creating a pre-release tag like x.y.za0 or x.y.zrc1. The version will be automatically adjusted in setup.py and when the code calls get_version() in _version.py.

edit: Now with tox for parallel testing of multiple python versions


scripts/test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+#put token in user's home dir
+printenv INDICO_API_TOKEN > ~/indico_api_token.txt
+#install additional test reqs and run smoke tests
+pip install requests-mock pytest
```
pastap, line 5: what's the reasoning to not use the tests_requires with setup.py in the Dockerfile?

codeship-steps.yml
```diff
@@ -1,8 +1,7 @@
-- type: parallel
-  steps:
-    - name: Python 2 Tests
-      service: python2
-      command: nosetests -sv .
-    - name: Python 3 Tests
-      service: python3
-      command: nosetests -sv .
+- name: test
+  service: indico-client
```
pastap, line 2: can we try and retain our parallel python tests to check 2 and 3? or are we deprecating python 2 support?

scripts/deploy.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+rm -rf /indico-client/dist
+pip install twine
+python setup.py sdist
+echo $TWINE_USERNAME
+echo $TWINE_REPOSITORY
+twine upload --repository testpypi /indico-client/dist/*
```
pastap, line 7: assuming you'll sub in $TWINE_REPOSITORY here? or is there a reason not to hardcode it?

scripts/deploy.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+rm -rf /indico-client/dist
+pip install twine
```
pastap, line 3: should this be ver-locked?

codeship-steps.yml
goatrocks: That is a good question, we absolutely could have a parallel run of the tests using python 2 or multiple versions of python, but are we planning to continue to support python 2 given the sunset date has passed over a year ago?

scripts/deploy.sh
goatrocks: I'll sub in $TWINE_RESPOSITORY or (more likely) remove it entirely, for now I just wanted to be really sure I didn't push anything anywhere besides testpypi.

scripts/test.sh
goatrocks: I wasn't sure if that would stay around when setup.py test was deprecated

scripts/deploy.sh
goatrocks: Most likely! I'll add that

codeship-steps.yml
goatrocks: Okay, per slack discussion it sounds like it's safe to drop python 2 support, but I did add in tox for parallel testing of multiple versions of 3, although I'll want to polish that some more

tox.Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:20.04
+
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y apt-transport-https
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y software-properties-common && add-apt-repository ppa:deadsnakes/ppa
+
+# ensure the base image has what we need
+#RUN DEBIAN_FRONTEND=noninteractive apt-get -y install build-essential software-properties-common
```
jkgenser, line 9: Do you need this anymore or want to keep for reference?

goatrocks: Hmm, no, don't need it any more as far as I can tell. I'll remove it so I don't leave commented out stuff!

PR #99, opened 2021-03-01
removing print line that contains error

Scott771:
Removed the print line from this code snippet returns list of ints, not submission objects, print line leads to exception

`submissions = client.call(
    WorkflowSubmission(files=["./path/to/sample.pdf"], workflow_id=52, submission=True)
)
print(submissions[0].id)`

NOTE: in second commit I updated my sphinx version to 3.5.1- we don't currently track sphinx version in the setup.py


Scott771:
I'm not familiar with the build error:
`Build Error! Build Error: services file error: please set either build or image option for service python3`

PR #107, opened 2021-04-08
Tiny pr to update force_render default setting

madisonmay:


docsrc/docextract_settings.rst
```diff
@@ -122,8 +122,8 @@ Force Render
 
 Setting::
 
-    "force_render": True
-    "force_render": False (default)
+    "force_render": True (default since IPA 4.7.0)
+    "force_render": False
```
fitzworkhub, line 126: do we want to be more explicit that this is the default pre IPA 4.8?

fitzworkhub (changes requested):

docsrc/docextract_settings.rst
madisonmay: sure! can't hurt

fitzworkhub (approved):

PR #108, opened 2021-04-23
[CAT-62] create python client for dataset user reporting

goatrocks:
This enables getting the dataset user metrics that are available in the API. I also pulled in @pastap 's older paginator PR -- I saw one of the comments about encapsulating the paginator function in the PagedRequest class but as that doesn't know about making http requests I don't think it would work there. I didn't include the parts of making submissions paged since I didn't want to impact something unrelated to this ticket.


indico/queries/usermetrics.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class _UserMetricsFilters(BaseType):
+    """Filters for fetching user metric data."""
+    date: str
+    user_id: str
+    user_email: str
+
+    def __init__(self, **kwargs):
+        self.date = kwargs.get('date', None).strftime('%Y-%m-%d') if kwargs.get('date', None) is not None else ""
```
pastap, line 25:
```suggestion
        self.date = kwargs['date'].strftime('%Y-%m-%d') if kwargs.get('date') else None
```
and can the default be None or does it have to be `""`

indico/queries/usermetrics.py
```diff
@@ -0,0 +1,113 @@
...
+    appRoles {
+      role
+      count
+    }
+  }
+}
+    """
+
+    def __init__(self, **kwargs):
+        date = kwargs.get('date', None)
```
pastap, line 56: why not explicitly call out `date` in the kwargs?
```suggestion
    def __init__(self, date=None):
```

indico/queries/usermetrics.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class _PagedUserSnapshots(BaseType):
+    """
+    Class to hold paged snapshot data to make parsing easier.
+    """
+    results: List[UserSnapshot]
+
+
+class _UserMetricsFilters(BaseType):
```
pastap, line 18: this is ok - but doesn't really encompass the full capabilities of the `filter` kwarg. See SubmissionFilter - basically would allow folks to build composite filters as part of their query

indico/queries/usermetrics.py
```diff
@@ -0,0 +1,113 @@
...
+    """Filters for fetching user metric data."""
+    date: str
+    user_id: str
+    user_email: str
```
pastap, line 22: Unfortunately can't link directly but [bullet point 7 under Graphql Schema](https://indicodata.atlassian.net/wiki/spaces/BE/pages/1352892435/User+Reports#Graphql-Schema)
user_id is a filter of type List[int] and user_email is a filter of type List[str]
so that folks can say give me userId=[1, 2, 3] - results for users where id=1 or 2 or 3

indico/types/user_metrics.py
```diff
@@ -0,0 +1,37 @@
...
+    count: int
+
+
+class UserDatasets(BaseType):
+    """Dataset Id and roles assigned to user in that dataset."""
+    dataset_id: int
+    role: str
+
+
+class UserSummaryUser(BaseType):
```
pastap, line 17: nit: is this a weird... name?

indico/queries/usermetrics.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Requests per-date detailed information about app users.
+    Args:
+        user_email (str): the email of a specific user.
+        user_id (int): the id of a specific user
+        date (datetime): specific day to query
+    """
+    query = """
+    query GetUserSnapshot($date: Date, $filters: UserReportFilter, $after: Int){
+  userSnapshot(date: $date, filters: $filters, after: $after){
```
pastap, line 77: may want to expose `limit` here for folks as an additional pagination option to expose?

indico/client/client.py
```diff
@@ -64,3 +64,16 @@ def call(self, request: Union[HTTPRequest, RequestChain]):
             return self._handle_request_chain(request)
         elif request and isinstance(request, HTTPRequest):
             return self._http.execute_request(request)
+
+    def paginate(self, request: PagedRequest):
```
pastap, line 68: if we're cool rolling with this - any hesitation with also including the change from the same PR to paginate ListSubmissions?

indico/queries/usermetrics.py
```diff
@@ -0,0 +1,113 @@
...
+        if date is not None:
+            super().__init__(self.query, variables={"date": date.strftime('%Y-%m-%d')})
+        else:
+            super().__init__(self.query)
+
+    def process_response(self, response) -> UserSummary:
+        return UserSummary(**super().process_response(response)["userSummary"])
+
+
+class GetUserSnapshots(PagedRequest):
```
pastap, line 66: will we also be exposing UserChangelog?

goatrocks:
Updated with feedback from @pastap :)

goatrocks:
Updated with the Changelog feature as well

PR #116, opened 2021-07-23
[CAT-103] FIX: Support gzip deserialization

thearchitector:
## Summary

Fixed an issue where `gzip`-compressed files retrieved by the HTTP client would fail deserialization if the inner content type was not unicode-encoded text. It just so happened to work before because we were coincidentally gzipping JSONs.

Breaking change as previously gzip-encoded JSONs fell back to string decoding, thus returned a string which happened to be correct. Results are now `bytes` since we can't know the inner MIME type without external libraries.

## Changelog
- Added `application/gzip` and `application/x-gzip` to the list of known MIME types.

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update


thearchitector:
i can make a JIRA if you'd like to track this, not sure where though

indico/http/serialization.py
```diff
@@ -85,5 +85,7 @@ def zip_serialization(content, charset=None):
         "image/png": image_serialization,
         "image/jpg": image_serialization,
         "application/zip": zip_serialization,
+        "application/x-gzip": raw_bytes,
```
pastap, line 88: This makes sense - but makes me wonder what the gzip option on `deserialize` is for... @goatrocks ? Any idea where we'd use that?

pastap (approved):

indico/http/serialization.py
thearchitector: it's gets automatically set if the file you're downloading ends in '.gz'. Tbh not great cause now we've kinda created an incorrect that stuff will magically work

thearchitector: occurred to me, do we want to keep doing the `.gz` reading? or should we only decompress if the mime type is correct?

pastap: I would say refer to mime-type and not to file name - but do we ever... set bad mimetypes? not sure... so to be safe, could just `if endswith(.gz) or mimetype == gzip`
that's just my guess tho

thearchitector: nginx sets the mime type, so i doubt we ever send bad ones. we'd have to manually override the response header which i highly doubt we do anywhere

goatrocks: What @thearchitector said is right, it gets set in `client.py` at line 131-138, and assumes that gzip content will end in .gz, but gzip content tends to come with the mime type set not the file extension so if it's enabled on a server this code doesn't work (hence the incorrect assumption)

I would lean towards removing the .gz assumption, but I'm also wary of removing things people may or may not be relying on, but then again I can't think of a situation in which we'd be returning a .gz file..

goatrocks (approved):

thearchitector:
@goatrocks @pastap yea so adding this def will break people who 1) dont know they are getting a gzip from IPA and 2) who assume gzips will decode to text automatically. the mime type will almost def be correct since nginx is setting it not us -- do we have a final vote on whether or not to remove the file extension logic and rely on mime type?

goatrocks:
> @goatrocks @pastap yea so adding this def will break people who 1) dont know they are getting a gzip from IPA and 2) who assume gzips will decode to text automatically. the mime type will almost def be correct since nginx is setting it not us -- do we have a final vote on whether or not to remove the file extension logic and rely on mime type?

I vote to remove it 👍

pastap:
> > @goatrocks @pastap yea so adding this def will break people who 1) dont know they are getting a gzip from IPA and 2) who assume gzips will decode to text automatically. the mime type will almost def be correct since nginx is setting it not us -- do we have a final vote on whether or not to remove the file extension logic and rely on mime type?
> 
> I vote to remove it

I follow the leader :D

PR #117, opened 2021-08-04
[CAT-102] byte stream support

goatrocks:
## Summary

This enables SDK users to manage their own file reading and upload to the backend by providing a map of filenames to  streams, so in addition to`files` on a `WorkflowSubmission`, you can add `io.BufferedIOBase` inheriting `streams` directly.

## Changelog (optional)
Modified WorkflowSubmission object to add optional streams
Modified Storage upload object to add optional streams
Modified client.py to handle the optional streams similar to files.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Checklist:
A quick checklist of things that should be done before the code is opened to review by others on the team.
Delete any that are not applicable.

- [ ] Documentation has been updated

### DIFF
--- a/indico/queries/storage.py
+++ b/indico/queries/storage.py
@@ -1,5 +1,6 @@
+import io
 import json
-from typing import List
+from typing import List, Dict
 from indico.client.request import HTTPMethod, HTTPRequest, RequestChain
 from indico.errors import IndicoRequestError, IndicoInputError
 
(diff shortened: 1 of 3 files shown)


indico/queries/storage.py
```diff
@@ -124,5 +131,5 @@ def process_response(self, uploaded_files: List[dict]) -> List[str]:
         return urls
 
 
-# Alias to ensure backwards compatability
+# Alias to ensure backwards compatibility
```
fitzworkhub, line 134: Compatability is how the brits write it ... like colour

indico/queries/workflow.py
```diff
@@ -266,18 +270,22 @@ def __init__(
         submission: bool = True,
         bundle: bool = False,
         result_version: str = None,
+        streams: Dict[str, io.BufferedIOBase] = None
```
fitzworkhub, line 273: Can this keys for this dict be file paths?

fitzworkhub (approved):

indico/queries/workflow.py
goatrocks: It could be if that's a requirement, and we could add some logic similar to the way we parse file names out for file paths, but ideally it would just be filename.ext

PR #122, opened 2021-10-08
Update datasets.py

madisonmay:
Default behavior for AddFiles should be to pass the autoprocess flag set to True


goatrocks (approved):
If the unit tests are working fine with this, lgtm

PR #127, opened 2021-11-29
Parallel push to GCR and Harbor Registries

ashmuck:
Adding codeship pipeline steps to push images to harbor as well as GCR. 
      Please note that if your Dockerfile has a reference to gcr.io, that build and push will not pass


codeship-services.yml
```diff
@@ -1,24 +1,22 @@
-indico-client-test:
```
sihrc, line 1: No real changes here right? Looks like it was just reordered

sihrc:
I don't see any changes in this PR aside from a new dockercfg. Where is this new dockercfg being used?

ashmuck:
No relevant changes made here

PR #136, opened 2022-02-04
[CAT-171] & [CAT-168] Fix exception handling for hibernation and decoding

goatrocks:
## Summary

Moved the error handling for 500 errors above the attempt to decode content so that an` IndicoDecodingError` doesn't get thrown before the server errors are checked.

Added a check for a 503 response code and throw an `IndicoHibernationError` if the retry-after header is set.

## Type of change

## Checklist:

### DIFF
--- a/indico/errors.py
+++ b/indico/errors.py
@@ -42,5 +42,6 @@ class IndicoAuthenticationFailed(IndicoError):
 
 class IndicoHibernationError(IndicoError):
     def __init__(self, after):
+        self.after = after
         super().__init__(f"Platform is currently hibernating. Wait {after} seconds and retry this request.")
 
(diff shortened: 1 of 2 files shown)


pastap (approved):
sick - dig the idea about suppressing 500 error information. a little nervous it'll be harder to debug if it's a server issue but ultimately, probably for the best. Future thought - wonder if it's one of those cases where a `debug` setting could be beneficial to expose the underlying error

goatrocks:
> sick - dig the idea about suppressing 500 error information. a little nervous it'll be harder to debug if it's a server issue but ultimately, probably for the best. Future thought - wonder if it's one of those cases where a `debug` setting could be beneficial to expose the underlying error

That's a good idea! I think for the 500s too we could potentially add a header from the backend that says "hey this 500 has more info/something to decode".

PR #165, opened 2022-05-19
update listworkflow query

nateshim-indico:
## Summary
Updates ListWorkflows query with updated schema


Scott771:
All looks good to me! But will defer approval to Meg

PR #180, opened 2022-07-08
[CAT-108] Enable autoprocessing by default for AddData

meghanhickey:
## Summary
Add the `autoprocess` flag to all `AddData` calls, and set the default the value to true. If `autoprocess` is true, no need for a separate `ProcessFiles` call. 

Update integration tests & docs to reflect this change


pastap:
Question - is this a breaking change in that the SDK will now do something different from what users expect - since their scripts generally AddtoDataset and then ProcessFiles? Esp if the first call launches auto-process, that could potentially launch 2 process pipelines for the same file - once through autoprocess and once through ProcessFiles

meghanhickey:
> Question - is this a breaking change in that the SDK will now do something different from what users expect - since their scripts generally AddtoDataset and then ProcessFiles? Esp if the first call launches auto-process, that could potentially launch 2 process pipelines for the same file - once through autoprocess and once through ProcessFiles

Yeah, good point. If @goatrocks agrees, perhaps we add the autoprocess flag but leave it off by default?

goatrocks:
> > Question - is this a breaking change in that the SDK will now do something different from what users expect - since their scripts generally AddtoDataset and then ProcessFiles? Esp if the first call launches auto-process, that could potentially launch 2 process pipelines for the same file - once through autoprocess and once through ProcessFiles
> 
> Yeah, good point. If @goatrocks agrees, perhaps we add the autoprocess flag but leave it off by default?

I agree, adding the flag and leaving it off by default would be the best approach

goatrocks (approved):

pastap (approved):
This suggestion is a bit extra, but you can always print/log a deprecation notice when someone tries to use the ProcessFiles function- and recommend autoprocess=True instead. As kind of a lead-up to making autoprocess default to True and maybe even removing the ProcessFiles function entirely
Not sure what the CAT process is for deprecation/pre-major-change notifications

meghanhickey:
> This suggestion is a bit extra, but you can always print/log a deprecation notice when someone tries to use the ProcessFiles function- and recommend autoprocess=True instead. As kind of a lead-up to making autoprocess default to True and maybe even removing the ProcessFiles function entirely Not sure what the CAT process is for deprecation/pre-major-change notifications

ooo good idea!

PR #188, opened 2022-08-03
FIX: Update docs

meghanhickey:
Updating docs to include current favicon and get rid of the separate submission InputFile type


fitzworkhub (approved):

meghanhickey:
tried deploying docs from the feature branch, and they look good, so merging to master.

PR #190, opened 2022-08-22
[CAT-322] and 320 and 321

fitzworkhub:
## Summary
add 

- ClassNames to modelGroup attribute in Workflows
- add filters attribute to workflowComponentLinks
- make labels an optional kwarg to AddLinkClassificationComponent
- Add after_link_id to AddModelGroupComponent

### DIFF
--- a/indico/types/workflow.py
+++ b/indico/types/workflow.py
@@ -69,7 +69,7 @@ class Workflow(BaseType):
 
 
 class LinkedLabelStrategy(Enum):
-    BY_ROW = (0,)
+    BY_ROW = 0,
     BY_KEY = 1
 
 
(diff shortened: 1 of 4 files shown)


indico/types/workflow.py
```diff
@@ -44,28 +46,41 @@ class Workflow(BaseType):
...
-                    and component.model_group.name == name)
+        return next(
+            component
+            for component in self.components
+            if hasattr(component, "model_group") and component.model_group.name == name
+        )
 
 
 class LinkedLabelStrategy(Enum):
-    BY_ROW = 0,
```
meghanhickey, line 63: we don't want to change this line-- should still be `BY_ROW = 0`

fitzworkhub: Not sure how that happened, reverted back to old version

meghanhickey (approved):

PR #191, opened 2022-08-26
[CAT-302] add code sample to create a classifier -> extraction workflow

fitzworkhub:


meghanhickey (changes requested):

meghanhickey (approved):
changes look good!

PR #200, opened 2022-11-14
feat: update storage url scheme

haizhou:
Update indico url scheme for blob

### DIFF
--- a/indico/queries/storage.py
+++ b/indico/queries/storage.py
@@ -2,12 +2,13 @@ import concurrent.futures
 import io
 import json
 from typing import Dict, List
-from urllib.request import Request
 
 from indico.client.request import GraphQLRequest, HTTPMethod, HTTPRequest, RequestChain
 from indico.errors import IndicoInputError, IndicoRequestError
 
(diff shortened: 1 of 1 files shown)


goatrocks:
does this go with #195 ?

haizhou:
> does this go with #195 ?

Yes. I think #195 is an additive feature, but this PR is a breaking change. I was also gonna ask how we manage client sdk version updates.

sihrc:
> > > does this go with #195 ?
> > 
> > 
> > Yes. I think #195 is an additive feature, but this PR is a breaking change. I was also gonna ask how we manage client sdk version updates.
> 
> We can create pre-release versions from a branch using github releases -- the version of the client SDK matches the minimum version of the platform. It sounds like this would go in a 6.0SDK -- and perhaps it also needs to get changed in the C# and Java SDKs as well?

tl;dr, we're not allowing breaking changes, so we're making the API compatible. We can move these changes into the other changes for blob storage.

haizhou:
Updated the merge base to #195

PR #203, opened 2023-01-04
[CAT-502] ADD: support for starting an integration

meghanhickey:
Add mutation and tests for starting an integration.

Note that in order to start an integration, the workflow must be runnable, and in order for the wf to be runnable, it needs an output node. The SDK doesn't yet have an official query for adding an output node. To facilitate testing, I added a hacky graphql call to add an output node to the org annotate workflow fixture. Once the SDK officially supports adding output nodes, we should replace this hacky call.


tests/integration/data/datasets.py
```diff
@@ -143,8 +143,28 @@ def org_annotate_dataset(indico):
 def org_annotate_workflow(indico, org_annotate_dataset: Dataset):
     client = IndicoClient()
     workflowreq = CreateWorkflow(dataset_id=org_annotate_dataset.id, name=f"OrgAnnotate-test-{int(time.time())}")
-    response = client.call(workflowreq)
-
+    wf = client.call(workflowreq)
+
+    # TODO: Replace this with AddComponent once it supports outputs
+    # currently on dev, json output component is blueprint 16
+    blueprint_id=16
```
pastap, line 150: blueprintId is optional for certain component types. rather than hard-coding, you can just specify component_type in your config as you already did and this should work

pastap: or actually, you should just be able to use _AddWorkflowComponent()

tests/integration/data/datasets.py
```diff
@@ -168,3 +188,27 @@ def org_annotate_model_group(indico, org_annotate_dataset: Dataset, org_annotate
     mg = client.call(GetModelGroup(id=component.model_group.id, wait=True))
 
     return mg
+
+@pytest.fixture(scope="module")
+def org_annotate_exchange_integration(org_annotate_workflow: Workflow) -> Integration:
+    client = IndicoClient()
+    creds = {
+        "clientId": os.getenv("EXCH_CLIENT_ID"),
```
pastap, line 196: are there some repo docs somewhere that specify that this env var should be set? there's also the pytest [skipif option](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1) which will check a conditional before deciding if the test should run. so for example skipif(os.getenv(CLIENT_ID) is None)

meghanhickey: > are there some repo docs somewhere that specify that this env var should be set?

not yet. the integration tests aren't in great shape-- we don't have any documentation for them, and they're not run during CI/CD because a significant portion time out. @goatrocks perhaps we should think about cleaning up and documenting these integration tests in the near future?

tests/integration/data/datasets.py
meghanhickey: when does the blueprint id need to be specified?

pastap: CUSTOM_MODEL and EXTERNAL_MODEL component types require it. optional otherwise (for backwards-compat with 5.X)

meghanhickey: :+1: dope, thanks for the info

pastap (approved):

PR #208, opened 2023-01-10
[CAT-511] add instructions for running tests

fitzworkhub:
## Summary
Add instructions for running tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ model_groups = response["model_groups"]["model_groups"]
 
 To run the tests associated with this repo perform the following:
 ### Prerequisite
-Set the `INDICO_API_TOKEN` environment variable to the token of the environment you need to run tests against.
+Set the `INDICO_HOST` environment variable to the address of the environment you'd like to run the integration tests against, and set `INDICO_API_TOKEN` to the token of the environment you need to run tests against. For more information about how to obtain a token, visit [the documentation](https://docs.indicodata.ai/articles/#!common-questions-publication/how-do-i-get-started-developing-with-the-indico-api/q/API%2520token/qid/3328/qp/1). 
 `export INDICO_API_TOKEN=<api_token>
 
 ### Running the tests
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ Set the `INDICO_API_TOKEN` environment variable to the token of the environment 
 `pip3 install pytest`
 5. Run tests
 `pytest -sv --host <indico_host> tests/`
-    * Note: you can run more specific tests by providing the path to the test e.g. ``pytest -sv --host <indico_host> tests/integration/test_base_client.py`
+    * Only run unit tests `pytest -sv --host <indico_host> tests/unit/`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +70,22 @@ response = client.call(GraphQLRequest(
...
+2. Activate the virtual environment
+`source venv/bin/activate`
+3. Install the client
+`python3 setup.py install`
+4. Install pytest
+`pip3 install pytest`
+5. Run tests
+`pytest -sv tests/`
+
+Note, you can be more specfic about tests you run by selecting a specific test folder or file.
```
meghanhickey, line 87: It'd be nice to have a subsection about running integrations tests specifically-- that you need to get an api token, how to set and change the cluster that the integration tests are pointed at

meghanhickey:
thanks for doing this!

fitzworkhub:
Totally forgot that was part of it.

README.md
fitzworkhub: Updated this

README.md
```diff
@@ -70,6 +70,26 @@ response = client.call(GraphQLRequest(
 model_groups = response["model_groups"]["model_groups"]
 ```
 
+# Testing the SDK
+
+To run the tests associated with this repo perform the following:
+### Prerequisite
+Set the `INDICO_API_TOKEN` environment variable to the token of the environment you need to run tests against.
```
meghanhickey, line 77:
```suggestion
Set the `INDICO_HOST` environment variable to the address of the environment you'd like to run the integration tests against, and set `INDICO_API_TOKEN` to the token of the environment you need to run tests against. For more information about how to obtain a token, visit [the documentation](https://docs.indicodata.ai/articles/#!common-questions-publication/how-do-i-get-started-developing-with-the-indico-api/q/API%2520token/qid/3328/qp/1). 
```

fitzworkhub: `INDICO_HOST` is not used in the tests. The tests derive the host from the `--host` cli arg when calling pytest. Made a point here: https://github.com/IndicoDataSolutions/indico-client-python/blob/5a4c85a121b3d0a26a8d4e3550603fa3cac3c540/README.md?plain=1#L90

meghanhickey: You can actually do either -- if you pass the `--host` argument to `pytest`, it automatically sets the `INDICO_HOST` environment variable: https://github.com/IndicoDataSolutions/indico-client-python/blob/master/tests/integration/conftest.py#L17

I think it's a little confusing that you can set it either way-- it makes most sense to me to remove the `pytest` option and require folks to set the host and the token via env. but *shrugs* it doesn't matter that much for the purposes of this PR.

meghanhickey (approved):

README.md
fitzworkhub: Even if you don't pass `--host` the default setting will cause it to default to `dev.indico.io`. I guess question for @goatrocks, sboudl we just remove this and rely on the environment variable to be available? Or was this potentially necessary for CI reason?

PR #213, opened 2023-01-20
[CAT-515] FIX: force decompress when deserializing if url ends in gz

meghanhickey:
In 5.x, the storage object returned by Document Extraction has a content type of `application/x-gzip`, but in 6.0 the content type is `binary/octet-stream`.

Since we can't always tell by the content type if the response needs to be decompressed, check the request URL and decompress the response if need be.

### DIFF
--- a/indico/http/client.py
+++ b/indico/http/client.py
@@ -1,6 +1,5 @@
 import http.cookiejar
 import logging
-from collections import defaultdict
 from contextlib import contextmanager
 from copy import deepcopy
 from pathlib import Path
@@ -149,15 +148,8 @@ class HTTPClient:
                 **new_kwargs,
             )
 
-        # code, api_response =
-        url_parts = path.split(".")
-        json = False
-        if len(url_parts) > 1 and (url_parts[-1] == "json" or url_parts[-2] == "json"):
-            json = True
-
-        decompress = False
-        if len(url_parts) > 1 and (url_parts[-1] == "gz"):
-            decompress = True
+        json: bool = ".json" in Path(path).suffixes
+        decompress: bool = Path(path).suffix == ".gz"
 
         # If auth expired refresh
         if response.status_code == 401 and not _refreshed:
--- a/indico/http/client.py
+++ b/indico/http/client.py
@@ -155,6 +155,10 @@ class HTTPClient:
         if len(url_parts) > 1 and (url_parts[-1] == "json" or url_parts[-2] == "json"):
             json = True
 
+        decompress = False
+        if len(url_parts) > 1 and (url_parts[-1] == "gz"):
+            decompress = True
+
         # If auth expired refresh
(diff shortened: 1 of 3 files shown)


fitzworkhub (commented):
Do we know why it's unclear why some need to be compressed and others don't?

meghanhickey:
> Do we know why it's unclear why some need to be compressed and others don't?

The issue is that the content type header for gzip files changed from 5.x to 6.x, so we can no longer rely on that header alone to tell if we need to decompress the file. I filed a bug on the PS board for the header change: https://indicodata.atlassian.net/browse/DEV-10316

indico/http/client.py
```diff
@@ -155,6 +155,10 @@ def _make_request(
         if len(url_parts) > 1 and (url_parts[-1] == "json" or url_parts[-2] == "json"):
             json = True
 
+        decompress = False
+        if len(url_parts) > 1 and (url_parts[-1] == "gz"):
+            decompress = True
+
```
sihrc, line 161: decompress = len(url_parts) > 1 and (url_parts[-1] == "gz")

path.split is typically better handled by pathlib to get an extension
`decompress: bool = Path(path).suffix == "gz"`

indico/http/serialization.py
```diff
@@ -48,6 +44,10 @@ def deserialize(response, force_json=False):
         )
 
 
+
+def raw_bytes(content, *args, **kwargs):
+    return content
```
sihrc, line 49: Where is this used?

tests/integration/conftest.py
```diff
@@ -9,7 +9,7 @@
 
 def pytest_addoption(parser):
     parser.addoption(
-        "--host", action="store", default="dev.indico.io", help="indico ipa host"
+        "--host", action="store", default="dev-ci.us-east-2.indico-dev.indico.io", help="indico ipa host"
```
sihrc, line 12: In terms of backwards compatibility, DEVOPS should just setup dev.indico.io CNAME to dev-ci.us-east-2.indico-dev.indico.io

indico/http/serialization.py
meghanhickey: https://github.com/IndicoDataSolutions/indico-client-python/pull/213/files#diff-ef5fa222e2bc2fb27c766b726185367b39ec9ef84ca3ac887ebec44f0b2dc7fcR74

tests/integration/conftest.py
sihrc: @meghanhickey sorry :( I think there's also an issue with the SSL cert not accepting dev.indico.io after all. We might just need to change these.

goatrocks (approved):
are there any changes we should make to 6.x to have the right content type come back for users who may not update their SDKs as quickly?

PR #215, opened 2023-01-31
[CAT-525] ADD: post processing submission status

meghanhickey:


fitzworkhub (commented):
Are there any if statements that use these enums? I had to make some changes in the C# library here: https://github.com/IndicoDataSolutions/indico-client-csharp/blob/adb001d18a6180b219c7d960eb08a7ff701c8471/IndicoV2/Extensions/SubmissionResult/SubmissionResultAwaiter.cs#L33

goatrocks (approved):

meghanhickey:
> Are there any if statements that use these enums? I had to make some changes in the C# library here: https://github.com/IndicoDataSolutions/indico-client-csharp/blob/adb001d18a6180b219c7d960eb08a7ff701c8471/IndicoV2/Extensions/SubmissionResult/SubmissionResultAwaiter.cs#L33

Thanks for pointing that out! I updated the equivalent function in the python sdk

PR #230, opened 2023-05-16
Updates for IPA 6.x

meghanhickey:


indico/queries/workflow_components.py
```diff
@@ -444,3 +444,5 @@ def process_response(self, response) -> Workflow:
         return Workflow(
             **super().process_response(response)["deleteWorkflowComponent"]["workflow"]
         )
+
+ # TODO Add mutation to add output node
```
goatrocks, line 448: should probably resolve this TODO or add a ticket somewhere

goatrocks (approved):

indico/queries/workflow_components.py
meghanhickey: In the words of Cousin Greg from Succession: if it is so be said, so it be, and so it is.

https://indicodata.atlassian.net/browse/CAT-635

we do still support adding an output node-- it would just be nice to make it its own lil class so the API is consistent for all components.

PR #231, opened 2023-05-17
[CAT-633] FIX: document extraction image test

meghanhickey:
The test for retrieving images from document extraction OCR output has been failing since 5.x. As of 5.x, the default OCR config no longer includes returning the image link. This PR updates the test to use the simple config, which does include image links.


fitzworkhub (approved):

goatrocks (approved):
lgtm

PR #250, opened 2023-07-14
[CAT-707] Add call to register custom blueprints

meghanhickey:
Add a single class to register different types of custom blueprints.

Open question: do we need to do more thorough type-checking of the config dict at the SDK level, or is it ok to let the platform (fog) handle that?


pastap:
> Add a single class to register different types of custom blueprints.
> 
> Open question: do we need to do more thorough type-checking of the config dict at the SDK level, or is it ok to let the platform (fog) handle that?

fair question. definitely leave the decision to you and micah but my 2 cents is that the fog API should ideally give you back very good details about what is wrong with the config, thus you may not want to make redundant checks and instead surface the BE error messages directly and cleanly

Additionally, and there are 2 schools of thought:
 If you keep the sdk model as simple as possible, it makes it easier for BE to introduce additions without SDK changes required
 But if you keep the sdk model very complex, it could save us from making backwards-incompat changes to the config parser in the BE

I defer to y'all :)

goatrocks:
> Add a single class to register different types of custom blueprints.
> 
> Open question: do we need to do more thorough type-checking of the config dict at the SDK level, or is it ok to let the platform (fog) handle that?

I agree with astha that most of the time the back end should surface more complex config issues, but I also think some simple checks here for values we know are always required is a good idea.

indico/queries/custom_blueprint.py
```diff
@@ -0,0 +1,98 @@
...
+            if not tags:
+                none_vals.append("tags")
+            raise IndicoInputError(f"The following arguments cannot be None: {','.join(none_vals)}")
+        if not all([isinstance(t, str) for t in tags]):
+            raise IndicoInputError("'tags' must be a list of strings")
+        if not all(config.values()):
+            raise IndicoInputError("values in the 'config' dict must not be None")
+        if not isinstance(component_family, ComponentFamily):
+            raise IndicoInputError("'component_family' must be of type indico.types.workflows.ComponentFamily")
+        if component_family.name not in ["OUTPUT", "FILTER", "MODEL"]:
```
goatrocks, line 75: this is the only one that we might want to leave out, in case the BE adds a new enum (maybe just check it's not none)?

goatrocks: That makes sense, especially #2! For ease of code maintenance, is it possible to keep a supported component family list in the file with the enum?

meghanhickey: good idea!

indico/queries/custom_blueprint.py
```diff
@@ -0,0 +1,136 @@
...
+    Args:
+        component_family (ComponentFamily): family this blueprint belongs to; supported families are Output, Filter, and Model
+        name (str): Name of the blueprint
+        description (str): Description of this blueprint
+        config (dict): Blueprint configuration options
+        tags: (list[str]): List of tags associated with this blueprint
+
+    Options:
+        footer (str): Footnote for this blueprint's description
+        icon (str): Name or storage location of image to use when displaying this blueprint in the gallery
```
fitzworkhub, line 28: is the name the local path to the image?

tests/integration/queries/test_custom_blueprint.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    bp = client.call(bpreq)
+    print(bp.__dict__)
+    assert bp.id
+
+def test_register_blueprint_bad_config(indico):
+    with pytest.raises(IndicoInputError):
+        RegisterCustomBlueprint(
+            component_family=ComponentFamily.OUTPUT, 
+            name="Meowtput",
```
fitzworkhub, line 31: nice

fitzworkhub (approved):

goatrocks (approved):

indico/queries/custom_blueprint.py
meghanhickey: no, there are a small subset of icons that come automatically with each cluster, where you only have specify the name. There's a bunch more info here: https://developer.indicodata.ai/docs/registering-a-custom-blueprint#selecting-an-icon

PR #278, opened 2023-11-06
[CAT-814] Clean up documentation for Readme

meghanhickey:
## Summary
- remove example rst files, since we want to put examples in recipes
- categorize doc pages as one of "classes" or "types" and sort them into folders
- remove specific markdown links that won't work in Readme, and edit other so they will work

### DIFF
--- a/indico/config/config.py
+++ b/indico/config/config.py
@@ -11,9 +11,9 @@ class IndicoConfig:
     Configuration for the IndicoClient.
 
     Args:
-        host= (str, optional): Indico Platform hostname (ie indico.myco.com)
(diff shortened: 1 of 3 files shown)


docsrc/index.rst
```diff
@@ -1,21 +1,7 @@
 ******************************
-indicoio Python Client Library
+Indico Python Client Library
```
goatrocks, line 2: Should we add any intro or guidance on using the guide here? Probably worth asking technical writing?

goatrocks (approved):

docsrc/index.rst
meghanhickey: what sort of stuff do you think we should add? like information on how to install the sdk?

PR #280, opened 2023-11-16
[CAT-837] Publish docs to readme

meghanhickey:


goatrocks:
Before we merge this we need to add the redirect/notice to the existing docs, and disable the github pages action that publishes it. we should get the info from technical writing for the verbiage around the redirect

goatrocks (approved):

PR #282, opened 2023-12-20
alpha async support

pastap:
Some tests run - more requests = more speedup!
image

There's a lot to test:
 - token refreshing
 - unit tests
 - pagination
 - weird cookie manipulation (works on defaults, I'm sure we'll wanna get some type of parity with the normal req standard)
 - CTX manager for http client so folks don't need to cleanup themselves

Keeping this in draft if we ever need it, I'll be attempting to incorporate it into our integration tests to get faster speedup of data setup, etc


goatrocks:
Really cool! We've been talking about working on an async client in Q1 + possibly revamping the way the client works.. we were considering using async httpx because https://github.com/mirumee/ariadne-codegen uses that (it also has retry + gzip decompression built in). it might be interesting to compare aiohttp to an async httpx version. 🤔

pastap:
> Really cool! We've been talking about working on an async client in Q1 + possibly revamping the way the client works.. we were considering using async httpx because https://github.com/mirumee/ariadne-codegen uses that (it also has retry + gzip decompression built in). it might be interesting to compare aiohttp to an async httpx version. 🤔

oh very very cool! sounds neat - hopefully it'll be a super easy swap in - looks like all the integration test fixtures are sync so I'm gonna port them to be async and then when the async client comes in, we'll swap in for my very ugly hacky class :D I know there are some perf benefits with aiohttp but I doubt it's anything insane that our clients would care about - httpx seems like an awesome way to go

thearchitector:
ive done some stuff in both -- `aiohttp` is def more performant/designed for highly concurrent workloads, but `httpx` has support for newer stuff like http/2, [anyio](https://anyio.readthedocs.io/en/stable/index.html), and direct-asgi-app-calling, which could appeal to larger customer base (though for the last one im not sure how useful that'd be / how many customers run client things within a python app vs. just as scripts). 

imo anyio as a paradigm is super neat, mainly in that is supports the [trio](https://trio.readthedocs.io/en/stable/)-style concurrency model (which is subjectively the correct way to do python async). so i personally would be extremely happy about the future py client supporting it.

re speed, there are some [workarounds](https://github.com/encode/httpx/issues/838#issuecomment-1291224189) for at least some perf degradations, or at least there were the last time i checked. not sure if they're still relevant

pastap:
Ran the integration tests I added
image

indico/client/client.py
```diff
@@ -80,3 +87,111 @@ def paginate(self, request: PagedRequest):
...
+                response = await self._handle_request_chain(request)
+                chain.previous = response
+
+        if chain.result:
+            return chain.result
+        return response
+
+    async def get_ipa_version(self):
+        if not self._created:
+            raise Exception("Please .create() your client")
```
goatrocks, line 156: Could we raise an IndicoError for this?

indico/http/client.py
```diff
@@ -186,3 +194,165 @@ def _make_request(
...
+    """
+
+    def __init__(self, config: Optional[IndicoConfig] = None):
+        """
+        Config options specific to aiohttp
+        unsafe - allows interacting with IP urls
+        """
+        self.config = config or IndicoConfig()
+        self.base_url = f"{self.config.protocol}://{self.config.host}"
+        unsafe = config.verify_ssl
```
goatrocks, line 212: Would we want to tie this to verify_ssl? Unsafe seems to allow cookies from IP addresses, but that would be slightly different behavior from allowing self signed ssl.

goatrocks (approved):

indico/http/client.py
```diff
@@ -186,3 +194,165 @@ def _make_request(
...
+
+        self.request_session = aiohttp.ClientSession()
```
meghanhickey, line 214: thoughts on using httpx here instead of aiohttp? mainly because we're using httpx for flurry and insights

nateshim-indico (dismissed):

indico/http/client.py
pastap: maybe one concern is that httpx needs python 3.8+ and we require working in py 3.7?

goatrocks: We dropped 3.7 since it's end of life https://devguide.python.org/versions/

pastap: nice! well then I'll see if I can get httpsx working within a short period of time

indico/http/client.py
pastap: great catch - looks like I use this to make requests anyway, not sure why I added this unless it was my dumb attempt at the cookie stuff - will remove!

goatrocks: All good! Nothing dumb about it, just part of the process 😅

goatrocks (approved):

PR #294, opened 2024-02-28
update submission filter

nateshim-indico:


indico/queries/submission.py
```diff
@@ -33,57 +33,95 @@ class ListSubmissions(PagedRequest):
...
+                    changes
+                }
+                retries {
+                    id
+                    submissionId
+                    previousErrors
+                    previousStatus
+                    retryErrors
+                }
+                AutoReviewLoaded
```
pastap, line 104: I recommend dropping this one - the attr is "private" and old as hell

indico/queries/submission.py
```diff
@@ -33,57 +33,95 @@ class ListSubmissions(PagedRequest):
...
+                ocrEngine
```
pastap, line 105: this one as well - I think it'll cause some pretty big performance slowness to include this

indico/queries/submission.py
```diff
@@ -33,57 +33,95 @@ class ListSubmissions(PagedRequest):
...
+                    id
+                    submissionId
+                    createdAt
+                    createdBy
+                    startedAt
+                    completedAt
+                    rejected
+                    reviewType
+                    notes
+                    changes
```
pastap, line 116: I'm conflicted about including `changes` - the payload could be massive so maybe we leave this off for now and now/future add a different `GetReviews` query for submissions that include changes

indico/types/filetype.py
```diff
@@ -0,0 +1,23 @@
+from enum import Enum
+
+class FileType(Enum):
```
pastap, line 3: don't get me wrong, I love an enum, but the SDK does become a little more brittle. for example, if we introduce zip support in the platform, a user will need to upgrade their SDK version whereas before, they'd be able to create a zip sub and retrieve it in queries without any problems

although... to be fair, I'm not sure I see where you use this enum?

indico/types/submission.py
```diff
@@ -71,32 +77,53 @@ class Submission(BaseType):
...
+        updated_at (datetime): Datetime the submission was updated
+        created_by (int): Id of the user who created the submission
+        updated_by (int): Id of the user who updated the submission
+        completed_at (datetime): Datetime the submission reached a completed state
+        files_deleted (bool): Submission files have been deleted (True) or not deleted (False) from file store
+        input_files (List[SubmissionFile]): The SubmissionFiles for the Submission
         input_file (str): URL of the first input datafile within the Indico Platform.
-        input_filename (str): name of the first original file
+        input_filename (str): Name of the first original file
         result_file (str): URL of the result datafile within the Indico Platform
```
pastap, line 94: maybe time to update the docs here as well. `URL of the _latest_ result file for this submission` or something to that effect?

indico/types/filetype.py
nateshim-indico: whoops yeah I ended up using `str` instead for that reason but forget to remove the unused enum definition. Removing now

indico/queries/submission.py
nateshim-indico: Sounds good I'm opting for just removing changes since class `SubmissionReviews` does not include it by default, and will create a ticket for a `GetReviews` query to be able to access changes given a review id

nateshim-indico: Ticket here: https://indicodata.atlassian.net/browse/DEV-11438

pastap:
final request - can you please add this CODEOWNERS file to this repo as part of your PR? it'll start automatically including folks in PRs against master https://github.com/IndicoDataSolutions/elnino/blob/development/CODEOWNERS

examples/submission-filters.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+sub_filter = SubmissionFilter(retrieved=True, input_filename="property")
+submissions = client.call(ListSubmissions(filters=sub_filter))
+
+"""
+Example 4
+List all submissions that are created and updated within a certain date range
+"""
+date_filter = DateRangeFilter(filter_from="2022-01-01", filter_to="2023-01-01")
+sub_filter = SubmissionFilter(created_at=date_filter, updated_at=date_filter)
```
jacobmanderson, line 41: for completeness sake, an additional ListSubmissions call for this example please

indico/filters/__init__.py
```diff
@@ -59,33 +59,73 @@ def __init__(
...
 
+class DateRangeFilter(dict):
+    """
+    Create a Filter when querying for Submissions within a certain date range
+    Args:
+        filter_from (str): A valid string representation of a datetime for start date to filter
+        filter_to (str): A valid string representation of a datetime for end date to filter
+    """
+
+    def __init__(self, filter_from: str = None, filter_to: str = None):
```
jacobmanderson, line 70: nit: typing should be `str | None` if we default to None. However maybe typing in the sdk is different, if so no worries

jacobmanderson (approved):
minor comments, not super necessary to change

indico/filters/__init__.py
goatrocks: typing should match the lowest supported version of python, so should be `Union[ str, None ]` to match 3.9+

goatrocks (approved):

PR #295, opened 2024-02-29
[DEV-11433] use uploadbatched instead of upload document

nateshim-indico:

### DIFF
--- a/indico/queries/workflow.py
+++ b/indico/queries/workflow.py
@@ -5,7 +5,7 @@ import tempfile
 
 from indico.client.request import GraphQLRequest, RequestChain, Debouncer
 from indico.errors import IndicoError, IndicoInputError
-from indico.queries.storage import UploadDocument
+from indico.queries.storage import UploadDocument, UploadBatched
 from indico.types import Job, Submission, Workflow, SUBMISSION_RESULT_VERSIONS
 from indico.types.utils import cc_to_snake, snake_to_cc
 
@@ -58,9 +58,9 @@ class ListWorkflows(GraphQLRequest):
                                         id
                                       }
                                 }
-                            
+
                         }
-                
+
                     }
                   componentLinks{
                     id
@@ -69,7 +69,7 @@ class ListWorkflows(GraphQLRequest):
                     filters{
                       classes
                     }
-                    
+
                   }
                 }
             }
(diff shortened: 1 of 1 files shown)


indico/queries/workflow.py
```diff
@@ -346,7 +346,7 @@ def __init__(
 
     def requests(self):
         if self.files:
-            yield UploadDocument(files=self.files)
+            yield UploadBatched(files=self.files, batch_size=10)
```
pastap, line 349: should batch_size be passed in as a kwarg, default to 10? in case of really large files, even 10 would be too large

nateshim-indico: Yeah that makes sense I can make that change. To be honest I always thought it was more on the user to control the amount and size of files being submitted, and sdk was supposed to just be sufficient to handle 1 large submission

pastap: right the server can only handle 500mb of files uploaded in 1 req. if a user has 10 60mb files, even if the server can process that (which it can), this SDK call will fail. if we expose the option then the user could set like batch_size 5 and be on their merry way :)

pastap: :thinking: makes me think mentioning the 500mb limit might be worth doing in the docstring for the batch_size argument, otherwise it's not obvious when/how to use the batch param. like also more efficient for the user to _increase_ batch size if the have 50 really small files

pastap: > supposed to just be sufficient to handle 1 large submission

for sure. with bundle=True tho, the "1 large submission" could be... large AF

nateshim-indico: > 🤔 makes me think mentioning the 500mb limit might be worth doing in the docstring for the batch_size argument, otherwise it's not obvious when/how to use the batch param. like also more efficient for the user to _increase_ batch size if the have 50 really small files

gotcha i'll add that in as well

indico/queries/workflow.py
```diff
@@ -305,15 +309,16 @@ class WorkflowSubmission(RequestChain):
...
+        self,
+        workflow_id: int,
+        files: List[str] = None,
+        urls: List[str] = None,
+        submission: bool = True,
+        bundle: bool = False,
+        result_version: str = None,
+        streams: Dict[str, io.BufferedIOBase] = None,
+        text: str = "",
+        batch_size=10,
```
pastap, line 321: typing on batch_size

indico/queries/workflow.py
goatrocks: Eventually we might want to add a little logic to this feature to optimize the batch for users (try to calculate a batch size based on file sizes/counts, or take an overriden one). overall, Astha's feedback I agree with 100% tho :)

nateshim-indico: added batch_size as kwarg

goatrocks (approved):

pastap (approved):

indico/queries/workflow.py
pastap: dynamic would be sick in the future :)

goatrocks:
Approved, from my understanding this uses existing fields so there's no need to wait for the next platform version to merge it.

PR #298, opened 2024-03-14
[DEV-11456] make request chains wait appropriately for sync/async calls

nateshim-indico:
## Summary
Request chains with a wait argument use time.sleep() internally, rather than yielding a delay to the client to sleep synchronously or asynchronously as appropriate. Moved sleeping to client and making all wait calls yield a float or integer denoting sleep time

## Type of change

## Checklist:

### DIFF
--- a/indico/client/client.py
+++ b/indico/client/client.py
@@ -42,7 +42,6 @@ class IndicoClient:
     def _handle_request_chain(self, chain: RequestChain):
         response = None
         for request in chain.requests():
-            print(request)
             if isinstance(request, HTTPRequest):
                 response = self._http.execute_request(request)
                 chain.previous = response
@@ -144,7 +143,6 @@ class AsyncIndicoClient:
     async def _handle_request_chain(self, chain: RequestChain):
         response = None
         for request in chain.requests():
-            print(request)
             if isinstance(request, HTTPRequest):
                 response = await self._http.execute_request(request)
                 chain.previous = response
--- a/indico/client/client.py
+++ b/indico/client/client.py
@@ -2,7 +2,8 @@
 
 from typing import Union, Optional
 import urllib3
-
+import time
+import asyncio
 from indico.config import IndicoConfig
 from indico.errors import IndicoError
 from indico.http.client import HTTPClient, AIOHTTPClient
@@ -41,13 +42,15 @@ class IndicoClient:
     def _handle_request_chain(self, chain: RequestChain):
         response = None
         for request in chain.requests():
+            print(request)
             if isinstance(request, HTTPRequest):
                 response = self._http.execute_request(request)
                 chain.previous = response
             elif isinstance(request, RequestChain):
                 response = self._handle_request_chain(request)
                 chain.previous = response
-
+            elif isinstance(request, (float, int)):
+                time.sleep(request)
         if chain.result:
             return chain.result
         return response
@@ -141,13 +144,15 @@ class AsyncIndicoClient:
     async def _handle_request_chain(self, chain: RequestChain):
         response = None
         for request in chain.requests():
+            print(request)
             if isinstance(request, HTTPRequest):
                 response = await self._http.execute_request(request)
                 chain.previous = response
             elif isinstance(request, RequestChain):
                 response = await self._handle_request_chain(request)
                 chain.previous = response
-
+            elif isinstance(request, (float, int)):
+                asyncio.sleep(request)
         if chain.result:
             return chain.result
         return response
--- a/indico/client/client.py
+++ b/indico/client/client.py
@@ -14,7 +14,7 @@ from indico.client.request import (
 )
 from indico.config import IndicoConfig
 from indico.errors import IndicoError
-from indico.client.request import Debouncer
+from indico.client.request import Delay
 from indico.http.client import AIOHTTPClient, HTTPClient
 
 
@@ -51,8 +51,8 @@ class IndicoClient:
             elif isinstance(request, RequestChain):
                 response = self._handle_request_chain(request)
                 chain.previous = response
-            elif isinstance(request, Debouncer):
-                time.sleep(request.backoff())
+            elif isinstance(request, Delay):
+                time.sleep(request.seconds)
         if chain.result:
             return chain.result
         return response
@@ -152,8 +152,8 @@ class AsyncIndicoClient:
             elif isinstance(request, RequestChain):
                 response = await self._handle_request_chain(request)
                 chain.previous = response
-            elif isinstance(request, Debouncer):
-                await asyncio.sleep(request.backoff())
+            elif isinstance(request, Delay):
+                await asyncio.sleep(request.seconds)
         if chain.result:
             return chain.result
         return response
--- a/indico/client/client.py
+++ b/indico/client/client.py
@@ -14,6 +14,7 @@ from indico.client.request import (
 )
 from indico.config import IndicoConfig
 from indico.errors import IndicoError
+from indico.client.request import Debouncer
 from indico.http.client import AIOHTTPClient, HTTPClient
 
 
@@ -50,8 +51,8 @@ class IndicoClient:
             elif isinstance(request, RequestChain):
                 response = self._handle_request_chain(request)
                 chain.previous = response
-            elif isinstance(request, (float, int)):
-                time.sleep(request)
+            elif isinstance(request, Debouncer):
+                time.sleep(request.backoff())
         if chain.result:
             return chain.result
         return response
@@ -151,8 +152,8 @@ class AsyncIndicoClient:
             elif isinstance(request, RequestChain):
                 response = await self._handle_request_chain(request)
                 chain.previous = response
-            elif isinstance(request, (float, int)):
-                asyncio.sleep(request)
+            elif isinstance(request, Debouncer):
+                await asyncio.sleep(request.backoff())
         if chain.result:
             return chain.result
         return response
--- a/indico/client/client.py
+++ b/indico/client/client.py
@@ -1,18 +1,20 @@
 # -*- coding: utf-8 -*-
 
-from typing import Union, Optional
-import urllib3
-import time
 import asyncio
+import time
+from typing import Optional, Union
+
+import urllib3
+
+from indico.client.request import (
+    GraphQLRequest,
+    HTTPRequest,
+    PagedRequest,
+    RequestChain,
+)
 from indico.config import IndicoConfig
 from indico.errors import IndicoError
-from indico.http.client import HTTPClient, AIOHTTPClient
-from indico.client.request import (
-    HTTPRequest,
-    RequestChain,
-    PagedRequest,
-    GraphQLRequest,
-)
+from indico.http.client import AIOHTTPClient, HTTPClient
 
 
 class IndicoClient:
--- a/indico/client/request.py
+++ b/indico/client/request.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Union
 
 from indico.errors import IndicoRequestError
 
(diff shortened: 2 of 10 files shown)


indico/client/client.py
```diff
@@ -47,7 +50,8 @@ def _handle_request_chain(self, chain: RequestChain):
             elif isinstance(request, RequestChain):
                 response = self._handle_request_chain(request)
                 chain.previous = response
-
+            elif isinstance(request, (float, int)):
```
pastap, line 53: this might be... a little brittle if the return type is ever a float or int. Have you thought about using something like a "Backoff" class instead? really simple
```
class Backoff:
  def __init__(self, time):
    self.time = time

return Backoff(5)
```
and then you can
```
if isinstance(req, Backoff):
  sleep(Backoff.time)
```

pastap: or even just leveraging a function to call like
if isinstance(request, Callable): request()...

pastap: hell yeah, sounds good - for sleeps, could just... 0 debounce or something yeah - great call

indico/queries/datasets.py
```diff
@@ -215,6 +215,7 @@ def __init__(
         name: str,
         files: List[str],
         wait: bool = True,
+        max_wait_time: Tuple[int, float] = 5,
```
nateshim-indico, line 218: Is it okay to add here or should this be at bottom of kwarg list

indico/client/request.py
```diff
@@ -90,13 +90,13 @@ def requests(self):
 
 
 class Debouncer:
-    def __init__(self, max_timeout: int = 5):
+    def __init__(self, max_timeout: Tuple[int, float] = 5):
```
nateshim-indico, line 93: Added kwarg `max_wait_time` to all client calls that use `Debouncer` to allow access to change this variable.

indico/client/request.py
```diff
@@ -90,13 +90,13 @@ def requests(self):
...
         self.timeout = 0
         self.max_timeout = max_timeout or 5  # prevent None and 0
 
-    def backoff(self):
+    def backoff(self) -> Tuple[int, float]:
```
pastap, line 97: same Union comment here

indico/queries/datasets.py
pastap: good call - bottom of list - just in case we got some folks who blindly pass params

pastap: and same Union, not Tuple comment

indico/queries/export.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         frozen_labelset_ids: List[int] = None,
         combine_labels: LabelResolutionStrategy = LabelResolutionStrategy.ALL.name,
         file_info: bool = None,
-        anonymoous: bool = None,
+        anonymous: bool = None,
```
pastap, line 57: lmao - I'm hoping no one used this otherwise technically a backwards-incompat change. will let @goatrocks make the final call here

indico/queries/jobs.py
```diff
@@ -67,11 +68,17 @@ class JobStatus(RequestChain):
 
     previous: Job = None
 
-    def __init__(self, id: str, wait: bool = True, request_interval=0.2, timeout=None):
+    def __init__(
+        self,
+        id: str,
+        wait: bool = True,
+        timeout: Tuple[int, float] = None,
+        max_wait_time: Tuple[int, float] = 5,
```
pastap, line 76: what's the diff? (also Union)

pastap: also - docstring update?

indico/queries/model_groups/model_groups.py
```diff
@@ -27,16 +23,19 @@ class GetModelGroup(RequestChain):
         ModelGroup object
     """
 
-    def __init__(self, id: int, wait: bool = False):
+    def __init__(
+        self, id: int, wait: bool = False, max_wait_time: Tuple[int, float] = 5
+    ):
         self.id = id
         self.wait = wait
+        self.max_wait_time = max_wait_time
```
pastap, line 31: docstring update, Union instead of Tuple, etc

indico/queries/export.py
goatrocks: If we call it out clearly in the release notes, given this isn't something like the submission api... should be ok to fix 🐄  😨

indico/queries/jobs.py
nateshim-indico: timeout makes call throw `IndicoTimeoutError` after a specified period of time. `max_wait_time` denotes the max wait time per retry call until JobStatus is COMPLETE. 

Updated docstring to hopefully make the distinction clear

indico/queries/export.py
nateshim-indico: Found this library that support argument deprecation: https://docs.twisted.org/en/stable/api/twisted.python.deprecate.html

Don't know how worth it would be to add it for this

indico/queries/datasets.py
```diff
@@ -552,21 +559,22 @@ def __init__(
...
         self.wait = wait
+        self.max_wait_time = max_wait_time
 
     def requests(self):
         yield _ProcessFiles(self.dataset_id, self.datafile_ids)
-        debouncer = Debouncer()
         yield GetDatasetFileStatus(id=self.dataset_id)
         if self.wait:
             while not all(
                 f.status in ["PROCESSED", "FAILED"] for f in self.previous.files
```
pastap, line 574: dumb question - does self.previous pick up the last `yield` in which case it'd mistakenly pick up the debouncer value instead of getdatasetfilestatus?

indico/queries/submission.py
```diff
@@ -378,12 +378,14 @@ def __init__(
         check_status: str = None,
         wait: bool = False,
         timeout: Union[int, float] = 30,
+        max_wait_time: Tuple[int, float] = 5,
```
pastap, line 381: Union

indico/queries/datasets.py
```diff
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 
 import json
-import jsons
 import tempfile
 from pathlib import Path
-from typing import List, Union, Dict, Optional
+from typing import Dict, List, Optional, Tuple, Union
```
pastap, line 6: unneeded Tuple import

indico/queries/export.py
pastap: or just print a `warnings.warn` if someone uses the old arg?

goatrocks: warnings.warn is good -- we have a deprecation library already but unfortunately it doesn't support params (or it didn't when it was added). :(

indico/queries/jobs.py
goatrocks: `retry_time_interval` or `time_between_retries` would be a clearer name for this variable I think.

indico/queries/submission.py
nateshim-indico: smh mb * 5

indico/queries/datasets.py
nateshim-indico: just tested and it does not. I did not think of that tbh so good thing for me to be aware of going forward thanks

indico/queries/jobs.py
```diff
@@ -67,15 +69,22 @@ class JobStatus(RequestChain):
...
-    def __init__(self, id: str, wait: bool = True, request_interval=0.2, timeout=None):
+    def __init__(
+        self,
+        id: str,
+        wait: bool = True,
+        timeout: Union[int, float] = None,
+    ):
         self.id = id
         self.wait = wait
-        self.request_interval = request_interval
```
nateshim-indico, line 73: @goatrocks @pastap  Should we `warnings.warn` before removing?

indico/queries/export.py
nateshim-indico: added typo argument back and checks for it as discussed

indico/queries/jobs.py
nateshim-indico: changed `max_wait_time` to `request_interval` to match existing argument in `JobStatus` query

jacobmanderson (approved):

indico/queries/datasets.py
```diff
@@ -196,12 +196,17 @@ class CreateDataset(RequestChain):
...
+        files (List[str]): List of path names to the dataset files.
+        wait (bool, optional): Wait for the dataset to upload and finish. Defaults to True.
+        dataset_type (str, optional): Type of dataset to create [TEXT, DOCUMENT, IMAGE]. Defaults to TEXT.
+        from_local_images (bool, optional): Flag whether files are local images or not. Defaults to False.
+        image_filename_col (str, optional): Image filename column. Defaults to 'filename'.
+        batch_size (int, optional): Size of file batch to upload at a time. Defaults to 20.
+        ocr_engine (OcrEngine, optional): Specify an OCR engine [OMNIPAGE, READAPI, READAPI_V2, READAPI_TABLES_V1]. Defaults to None.
+        omnipage_ocr_options (OmnipageOcrOptionsInput, optional): If using Omnipage, specify Omnipage OCR options. Defaults to None.
+        read_api_ocr_options: (ReadApiOcrOptionsInput, optional): If using ReadAPI, specify ReadAPI OCR options. Defaults to None.
+        request_interval (int or float, optional): The maximum time in between retry calls when waiting. Defaults to 5.
```
goatrocks, line 209: bless your doc updates :)

indico/queries/datasets.py
```diff
@@ -538,9 +544,10 @@ class ProcessFiles(RequestChain):
     Process files associated with a dataset and add corresponding data to the dataset
 
     Args:
-        dataset_id (int): ID of the dataset
-        datafile_ids (List[str]): IDs of the datafiles to process
-        wait (bool): Block while polling for status of files
+        dataset_id (int): ID of the dataset.
+        datafile_ids (List[str]): IDs of the datafiles to process.
+        wait (bool, optional): Block while polling for status of files. Defaults to True.
+        request_interval (int or float, optional): The maximum time in between retry calls when waiting. Defaults to 5.
```
goatrocks, line 550: 5 seconds? Minutes?

indico/queries/export.py
goatrocks: Thank you Nate!

goatrocks (dismissed):
lgtm pending @mawelborn 's feedback 👍

mawelborn (changes requested):
Couple issues:

1. [Debouncing](https://underscorejs.org/#debounce) is a specific event-based programming technique that I was surprised to see pop up here. Reading the code, I see it's actually doing something related to backoff.
2. Regardless of the `request_interval` or `Debouncer.max_timeout` specified, the client will always sleep 1 second because that's where the `debouncer` starts, and `debouncer.backoff()` never gets called twice on a single instance.

As someone who creates request chains, but isn't overly familiar with the client codebase, I found this to be a bit confusing. If the desire is to avoid yielding a bare `float` for delays (understandable), my recommendation would be to replace it with a simple `Delay` dataclass to improve readability and fix the issue of not being able to yield a longer delay.

```python
@dataclass
class Delay:
    seconds: float
```

```python
class IndicoClient:
    ...

    def _handle_request_chain(self, chain: RequestChain):
        ...
        elif isinstance(request_or_delay, Delay):
            time.sleep(request_or_delay.seconds):
```

```python
class WaitUntilComplete(RequestChain):
    ...

    def requests(self) -> Iterator[GraphQLRequest | Delay]:
        yield GetSubmission(self.submission_id)
        while self.previous.status != "COMPLETE":
            yield Delay(seconds=8)
            yield GetSubmission(self.submission_id)
```

nateshim-indico:
Debouncer was used for CreateExport call (I don't know the history of why) and so we leveraged it throughout for consistency.

@goatrocks @pastap I'm thinking Debouncer is probably not needed for CreateExport and we can use simple Delay class throughout instead? I don't think it makes sense to have both especially if Debouncer will be used for a single call (CreateExport)

mawelborn (approved):
Beautiful!

goatrocks (approved):

PR #299, opened 2024-03-25
[DEV-11487] Remove deprecated methods 

nateshim-indico:
## Summary
Removing methods marked as deprecated in previous releases

## Checklist:


tests/integration/queries/test_workflow.py
```diff
@@ -515,12 +543,22 @@ def _new_dataset_for_updating(client):
...
     )
 
+    questionnaire = client.call(
+        GetQuestionnaire(
+            questionnaire.model_group_by_name(
+                classifier_name
+            ).model_group.questionnaire_id
+        )
+    )
+    while questionnaire.questions_status == "STARTED":
```
nateshim-indico, line 553: I think GetQuestionnaire should be a `RequestChain` with `wait` kwarg - previously CreateQuestionnaire handled the waiting but AddModelComponent does not wait on Questionnaire status which could cause confusion because a Questionnaire num_examples could be off when calling GetQuestionnaire right after adding model component.

goatrocks:
You'll want to write up some release notes that call out each function and their alternative as part of this work if that isn't done already :)

jacobmanderson (approved):
so much red

pastap (approved):
beautiful

tests/integration/queries/test_workflow.py
nateshim-indico: will make separate pr for this

PR #302, opened 2024-04-01
[DEV-11501] Update Model Group Settings 

nateshim-indico:
## Summary
Add endpoint to allow users to update model group training options 

## Type of change

## Checklist:

### DIFF
--- a/indico/queries/model_groups/model_groups.py
+++ b/indico/queries/model_groups/model_groups.py
@@ -7,7 +7,7 @@ from indico.client.request import Delay, GraphQLRequest, RequestChain
 from indico.errors import IndicoNotFound
 from indico.queries.workflow_components import AddModelGroupComponent
 from indico.types.jobs import Job
-from indico.types.model import Model
+from indico.types.model import Model, ModelOptions
 from indico.types.model_group import ModelGroup
 from indico.types.utils import cc_to_snake
 
@@ -235,4 +235,51 @@ class ModelGroupPredict(RequestChain):
 
         yield _ModelGroupPredict(
             model_id=self.model_id, data=self.data, predict_options=self.predict_options
+        )
+
+
+class UpdateModelGroupSettings(GraphQLRequest):
+    """
+    Updates an existing model group component in the platform.
+
+    Args:
+        model_group_id (int): the id of the model group to update settings
+        model_training_options (dict): model training options to use when training model
+    """
+
+    query = """
+        mutation updateModelGroup($modelGroupId: Int!, $modelTrainingOptions: JSONString) {
+            updateModelGroupSettings(
+                modelGroupId: $modelGroupId
+                modelTrainingOptions: $modelTrainingOptions
+            ) {
+                modelOptions {
+                    id
+                    modelTrainingOptions
+                }
+            }
+        }
+    """
+
+    def __init__(
+        self,
+        model_group_id: int,
+        model_training_options: dict = None,
+    ):
+        if model_training_options:
+            model_training_options = json.dumps(model_training_options)
+
+        super().__init__(
+            self.query,
+            variables={
+                "modelGroupId": model_group_id,
+                "modelTrainingOptions": model_training_options,
+            },
+        )
+
+    def process_response(self, response):
+        return ModelOptions(
+            **super().process_response(response)["updateModelGroupSettings"][
+                "modelOptions"
+            ]
         )
--- a/indico/types/model.py
+++ b/indico/types/model.py
@@ -29,3 +29,16 @@ class Model(BaseType):
     id: int
     status: str
     training_progress: TrainingProgress
+
+
+class ModelOptions(BaseType):
+    """
+    Model Options in the Indico Platform.
+
+    Attributes:
+        id (int): The model options id
+        model_training_options (str): JSONString representation of the model training options specified
+    """
+
+    id: int
+    model_training_options: str
--- a/tests/integration/queries/test_model_group.py
+++ b/tests/integration/queries/test_model_group.py
@@ -13,6 +13,7 @@ from indico.queries.model_groups import (
     GetModelGroup,
     GetTrainingModelWithProgress,
     ModelGroupPredict,
+    UpdateModelGroupSettings,
 )
 from indico.queries.model_groups.metrics import (
     AnnotationModelGroupMetrics,
@@ -95,7 +96,9 @@ def test_model_group_progress(
         )
     )
     time.sleep(1)
-    model: Model = client.call((GetTrainingModelWithProgress(id=mg.model_group_by_name(name).id)))
+    model: Model = client.call(
+        (GetTrainingModelWithProgress(id=mg.model_group_by_name(name).id))
+    )
 
     assert type(model) == Model
     assert model.status in ["CREATING", "TRAINING", "COMPLETE"]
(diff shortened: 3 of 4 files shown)


indico/queries/model_groups/model_groups.py
```diff
@@ -236,3 +236,50 @@ def requests(self):
...
         )
+
+
+class UpdateModelGroupSettings(GraphQLRequest):
+    """
+    Updates an existing model group component in the platform.
+
+    Args:
+        model_group_id (int): the id of the model group to update settings
+        model_training_options (dict): model training options to use when training model
```
pastap, line 247: may want to add predictoptions here as well - ie cover the whole API instead of just a subset of the arguments

indico/queries/model_groups/model_groups.py
```diff
@@ -236,3 +236,50 @@ def requests(self):
...
+                    modelTrainingOptions
+                }
+            }
+        }
+    """
+
+    def __init__(
+        self,
+        model_group_id: int,
+        model_training_options: dict = None,
```
pastap, line 267: typing for a default None should always include None in the option 
```suggestion
        model_training_options: dict | None = None,
```
also can't remember if the sdk is supposed to remain 3.8 compatible in which case you'll need to use the typing library (Optional[Dict[str, Any]]) instead

indico/types/model.py
```diff
@@ -29,3 +29,16 @@ class Model(BaseType):
...
+    """
+    Model Options in the Indico Platform.
+
+    Attributes:
+        id (int): The model options id
+        model_training_options (str): JSONString representation of the model training options specified
+    """
+
+    id: int
+    model_training_options: str
```
pastap, line 44: should this be dict or None? and should we deserialize json here?

tests/integration/queries/test_model_group.py
```diff
@@ -95,7 +96,9 @@ def test_model_group_progress(
         )
     )
     time.sleep(1)
-    model: Model = client.call((GetTrainingModelWithProgress(id=mg.model_group_by_name(name).id)))
+    model: Model = client.call(
+        (GetTrainingModelWithProgress(id=mg.model_group_by_name(name).id))
```
pastap, line 100: extra parens?

indico/types/model.py
nateshim-indico: making it dict or None is causing issues in BaseType: `TypeError: issubclass() arg 1 must be a class`

Opting to make it just dict for now

indico/queries/model_groups/model_groups.py
```diff
@@ -236,3 +236,104 @@ def requests(self):
...
         )
+
+
+class UpdateModelGroupSettings(GraphQLRequest):
+    """
+    Updates an existing model group component in the platform.
+
+    Args:
+        model_group_id (int): the id of the model group to update settings
+        model_training_options (dict, optional): model training options to use when training model. Defaults to None
```
fitzworkhub, line 247: Do we want to document the available model training options? Or at least link to them?

indico/queries/model_groups/model_groups.py
```diff
@@ -236,3 +236,104 @@ def requests(self):
...
+            Valid values include CLASSIFICATION, CLASSIFICATION_MULTIPLE, RATIONALIZED_CLASSIFICATION, REGRESSION, ANNOTATION, OBJECT_DETECTION, FORM_EXTRACTION, CLASSIFICATION_UNBUNDLING
+            Defaults to None
+        test_split (float, optional): denotes test split. Defaults to None
+        word_predictor_strength (str, optional): denotes word predictor strength
+            Valid values include STRONG, MODERATE, WEAK
+            Defaults to None
+    """
+
+    query = """
+        mutation updateModelGroup($modelGroupId: Int!, $modelTrainingOptions: JSONString) {
```
fitzworkhub, line 273: How come with only have 2 args for this query if there's like 12 variables?

pastap: same question here - think you're missing some mutation args

indico/queries/model_groups/model_groups.py
```diff
@@ -236,3 +236,104 @@ def requests(self):
...
+    def __init__(
+        self,
+        model_group_id: int,
+        model_training_options: Optional[Dict[str, Any]] = None,
+        predict_options: Optional[Dict[str, Any]] = None,
+        domain: str = None,
+        finetune: bool = None,
+        interlabeler_resolution: str = None,
+        make_predictions: bool = None,
+        roc_auc_averaging: str = None,
```
pastap, line 304: same comment again as before - None defaults need to be typed as optional

indico/queries/model_groups/model_groups.py
```diff
@@ -236,3 +236,104 @@ def requests(self):
...
+        model_group_id (int): the id of the model group to update settings
+        model_training_options (dict, optional): model training options to use when training model. Defaults to None
+        predict_options (dict, optional): predict options to use on model. Defaults to None
+        domain (str, optional): feature domain to use.
+            Valid values include STANDARD, STANDARD_V2, TOPICS, SENTIMENT, FINANCE, EMOTION, ENSEMBLE, FASTTEXT, UNSUPERVISEDSENTIMENT, IMAGE_V2, IMAGE_V3, IMAGE_V4, IMAGE_ENSEMBLE
+            Defaults to None
+        finetune (bool, optional): flag to use finetune or not in model training. Defaults to None
+        interlabeler_resolution (str, optional): denotes label resolution strategy.
+            Valid values include MAJORITY_VOTE_WITH_TIES, MAJORITY_VOTE_WITHOUT_TIES, UNANIMOUS, ALL
+            Defaults to None
```
pastap, line 255: these are deprecated so let's drop these

tests/integration/queries/test_model_group.py
```diff
@@ -206,3 +209,19 @@ def check_annotation_metrics(result):
     assert isinstance(result.model_level_metrics[0].macro_f1, float)
     assert isinstance(result.model_level_metrics[0].weighted_f1, float)
     assert result.retrain_for_metrics is False
+
+
+def test_update_model_group_settings(indico, org_annotate_model_group):
+    client = IndicoClient()
+    model_training_options = {"use_autolabeled_data": True, "use_partial_data": True}
```
goatrocks, line 216: we should test with the `ModelOptions` object

indico/types/model.py
```diff
@@ -29,3 +32,34 @@ class Model(BaseType):
     id: int
     status: str
     training_progress: TrainingProgress
+
+
+class ModelOptions(BaseType):
```
goatrocks, line 37: Should this be `ModelTrainingOptions` ?

jacobmanderson: i thought the same too but saw that it had an attribute model_training_options. if this is supposed to be model training options maybe we can update the typing in update_model_group_settings to make creating the model options a bit easier on the user side

tests/integration/queries/test_model_group.py
nateshim-indico: ModelOptions is just the return type model_training_options/predict_options are dict inputs

We test that the return type is ModelOptions and test that certain parameters are populated is that enough or should we cover other cases just to be safe?

goatrocks: I'm just easily confused haha. 

But, if it's possible, could we provide typed input for users?

nateshim-indico: We could but model_training_options and predict_options take in different inputs depending on what model type they're used on. We would probably be creating like 4 different types and adding logic to determine which type to accept. The documentation is updated to inform the user of which settings to use so maybe that's enough instead?

goatrocks: yeah, fair, maybe something to take for a future refactoring and we can add a readme guide for now?

nateshim-indico: ah yep ill make a ticket for that

PR #303, opened 2024-04-03
[DEV-11527] add: backwards compat static typing

thearchitector:
## Summary

type stubs and mypy compliance. also ran our standard set of pre-commit linters against everything

enables using the client in typed contexts (all of Insights) with in-editor sanity checks without doing nasty type casts or ignores:
```py
client: IndicoClient = IndicoClient(IndicoConfig(...))

# this is valid
subs: list[Submission] = client.call(ListSubmission(...))
# this is also valid
subs: Iterator[list[Submission]] = client.paginate(ListSubmission(...))

# this is not, and will throw a typing error 
subs: list[str] = client.call(ListSubmission(...))
```

also works for the async client, example in `examples/typed_calls.py`

---

enabled by turning http request classes into a generic that exposes its anticipated return type. for ex:
```py
class GetIPAVersion(GraphQLRequest[str]):
    query = """
        "query getIPAVersion {
            ipaVersion
        }
    """

    def __init__(self) -> None:
        super().__init__(self.query)

    def process_response(self, response: "Payload") -> str:
        version: str = super().parse_payload(response)["ipaVersion"]
        return version
```

---

All typing should be compatible with the versions we say we test against (3.8+), and should not cause any breaking changes for people who are not using static typing.

**verifications of that claim** -- i made no material changes to any of the existing tests, only the linting:

<details>
<summary>tox:</summary>

```
  py38: OK (4.98=setup[3.90]+cmd[0.55,0.53] seconds)
  py39: OK (5.01=setup[3.95]+cmd[0.46,0.60] seconds)
  py310: OK (4.73=setup[3.75]+cmd[0.46,0.52] seconds)
  py311: OK (5.03=setup[3.97]+cmd[0.46,0.60] seconds)
  py312: OK (5.09=setup[4.02]+cmd[0.49,0.58] seconds)
  py313: OK (5.02=setup[3.99]+cmd[0.55,0.48] seconds)
  congratulations :) (5.13 seconds)
```

</details>

<details>
<summary>mypy:</summary>

```
(venv) indico@XPS-13-9310:~/intake/indico-client-python$ mypy --config-file=pyproject.toml
Success: no issues found in 57 source files
```

</details>

<details>
<summary>integration tests:</summary>

these tests were already failing

```
FAILED tests/integration/queries/test_custom_blueprint.py::test_register_blueprint - indico.errors.IndicoRequestError: Status: 400, Error: A ComponentType.CUSTOM_OUTPUT blueprint called Meowtput with footer version 2 already exists.
FAILED tests/integration/queries/test_model_group.py::test_model_group_progress - indico.errors.IndicoRequestError: Status: 400, Error: 'NoneType' object is not subscriptable
FAILED tests/integration/queries/test_questionnaire.py::test_get_examples - assert 0 == 3
FAILED tests/integration/queries/test_questionnaire.py::test_add_labels - TypeError: AddLabels.__init__() got an unexpected keyword argument 'dataset_id'
FAILED tests/integration/queries/test_workflow.py::test_workflow_submission[_input0] - AttributeError: 'Submission' object has no attribute 'deleted'
FAILED tests/integration/queries/test_workflow.py::test_workflow_submission[_input1] - AttributeError: 'Submission' object has no attribute 'deleted'
```

vs. the run with my changes:
```
FAILED tests/integration/queries/test_custom_blueprint.py::test_register_blueprint - indico.errors.IndicoRequestError: Status: 400, Error: A ComponentType.CUSTOM_OUTPUT blueprint called Meowtput with footer version 2 already exists.
FAILED tests/integration/queries/test_model_group.py::test_model_group_progress - indico.errors.IndicoRequestError: Status: 400, Error: 'NoneType' object is not subscriptable
FAILED tests/integration/queries/test_questionnaire.py::test_get_examples - assert 0 == 3
FAILED tests/integration/queries/test_questionnaire.py::test_add_labels - TypeError: AddLabels.__init__() got an unexpected keyword argument 'dataset_id'
FAILED tests/integration/queries/test_workflow.py::test_workflow_submission[_input0] - AttributeError: 'Submission' object has no attribute 'deleted'
FAILED tests/integration/queries/test_workflow.py::test_workflow_submission[_input1] - AttributeError: 'Submission' object has no attribute 'deleted'
================================================================= 6 failed, 109 passed, 1 skipped in 3946.57s (1:05:46) =================================================================
```

</details>


indico/client/request.py
```diff
@@ -13,41 +21,55 @@ class HTTPMethod(Enum):
...
             raise IndicoRequestError(
                 error="\n".join(error["message"] for error in errors),
                 code=400,
                 extras=extras,
             )
-        return response["data"]
 
+        return raw_response["data"]
+
+    def process_response(self, response: "AnyDict") -> "ResponseType":
```
thearchitector, line 66: typed python does not permit overriding the typing of the signature in subclasses. we were doing that implicitly before by having this class return `Dict` and subclasses return their `BaseType`, etc. 

to avoid having to type every single use of `super().process_response` with a `cast()` or `type: ignore`, i split the function into a `parse_payload` typed to return `Any`.

practically, that means we're deprecating the use of `super().process_response` in any custom `GraphQLRequest` subclasses; i tried, but could not figure out a nice way to raise a deprecation error so folks know to switch over

indico/types/base.py
```diff
@@ -28,18 +37,22 @@ def valid_type(v):
     )
 
 
+# TODO: all the introspection here class makes the typing sus
```
goatrocks, line 40: I wonder if we can do away with this class at some point

thearchitector: moving to pydantic is my next pet project

indico/types/workflow.py
```diff
@@ -90,8 +92,8 @@ def __init__(
 
 
 class ComponentFamily(Enum):
-    MODEL = 0,
-    FILTER = 1,
-    TRANSFORMER = 2,
-    REVIEW = 3,
+    MODEL = (0,)
+    FILTER = (1,)
```
goatrocks, line 96: I don't remember why these are necessary 🤔 but I know they are?

indico/types/base.py
goatrocks: cooldown collab?

indico/client/client.py
```diff
@@ -202,3 +251,12 @@ async def paginate(self, request: PagedRequest):
         while request.has_next_page:
             r = await self._http.execute_request(request)
             yield r
+
+
+# class test(GraphQLRequest[int]):
```
jacobmanderson, line 256: are we good to remove this?

indico/types/workflow.py
thearchitector: yea i noticed that, and figured i best not touch it. black just added parens

jacobmanderson (dismissed):

indico/client/client.py
thearchitector: whoops, yea

indico/client/request.py
goatrocks: We've used https://deprecation.readthedocs.io/en/latest/  in the past for raising deprecation errors

thearchitector: yea, i just couldn't figure out a way to _only_ raise the error when invoking it via `super().process_response`

goatrocks: my kingdom for working deprecation in python

jacobmanderson (dismissed):

indico/client/request.py
```diff
@@ -1,8 +1,16 @@
...
+from typing import TYPE_CHECKING, Generic, TypeVar, cast
 
 from indico.errors import IndicoRequestError
 
+if TYPE_CHECKING:  # pragma: no cover
+    from typing import Any, Dict, Iterator, List, Optional, Union
+
+    from indico.typing import AnyDict
+
+ResponseType = TypeVar("ResponseType")
```
thearchitector, line 12: note: should this be `covariant=True`? that would enable stuff like this
```py
workflow: Workflow | None = None
workflow = client.call(GetWorkflow(...))
```
which is currently disallowed

not super clear to me if there are any negative repercussions of this

sihrc: Personally worked around this by not using None and raising an error if a workflow is expected. I find try except pattern to cover my cases usually.

Is this a new limitation in python 3.12? I don't think I've had this typing issue before.

thearchitector: not new, mypy 1.8 on Python 3.10

covariance also allows this, which I could see as more of an argument as to why it's useful
```py
datasets: list[Dataset] = []
datasets.extend(
  client.call(ListDatasets(...))
)
```
because std `extend` is typed as
```py
def extend(self, iterable: Iterable[_T])
```
(so Iterable vs. List)

goatrocks:
we can cut a pre-release and have some alpha/uat testing done here @nicholas-lockhart I think it would be worth it if folks have time to help test?

madisonmay:
we should get this merged relatively soon so it doesn't rot on the vine

thearchitector:
perhaps the best way to test this would be to switch out the client version used in the intake integration tests, see if they pass, and if so assert that means we're safe

indico/client/client.py
```diff
@@ -4,7 +4,7 @@
 import time
 from typing import TYPE_CHECKING, cast
 
-import urllib3
+import urllib3  # type: ignore
```
thearchitector, line 7: these won't gen type errors if you install the typing packages locally:

https://github.com/IndicoDataSolutions/indico-client-python/tree/elias/typing?tab=readme-ov-file#running-the-tests

this just doesn't use poetry so we cant use dependency groups 😢

indico/queries/datasets.py
```diff
@@ -88,7 +88,9 @@ def __init__(
             },
         )
 
-    def process_response(self, response: "Payload") -> "List[Dataset]":
+    def process_response(
+        self, response: "Payload", _: "Optional[List[str]]" = None
```
thearchitector, line 92: this and all the others don't throw an invalid signature override against `PagedRequest`?

thearchitector (commented):
@Sung96kim if you didn't see it already, i usually re-check type compliance and unit tests after rebasing with `docker compose run --rm tester`

indico/client/client.py
Sung96kim: actually planning on just getting the poetry pr in too so i type ignored it for now

indico/queries/datasets.py
Sung96kim: nope, the almighty mypy was satisfied

Sung96kim: `Success: no issues found in 59 source files`

thearchitector: have i forgotten or am i tripping? where is this invoked with 2 args where the second is an optional list

Sung96kim: for pagedrequest, it was added in the most recent pr

thearchitector: nested pagination sounds like a lot of really intuitive fun

Sung96kim (approved):
LGTM

indico/queries/datasets.py
Sung96kim: 🙃 almost on par with .bin files amirite

thearchitector (commented):
agree its good to go

image

PR #308, opened 2024-05-06
DEV-11576 Set autoprocess to True

madisonmay:


nicholas-lockhart:
What's the context for this change? As someone who is still learning how all of this works, it looks like we set autoprocess to false when adding files but set it to true when creating a dataset. To me, having this be the default as true makes sense from a (ignorant) user perspective because I would want files added to be processed without me doing an additional step. However, what was the reasoning behind the false boolean to begin with?

goatrocks (approved):

goatrocks:
@nicholas-lockhart It is an older feature that gets confusing, so changing the default makes sense. We changed the default for dataset creation previously

madisonmay:
I think it was originally false simply because we wanted new behavior with the auto process flag to default to existing behavior. 

It's the cause of a common issue where folks believe their dataset is stuck because the file has been downloaded but not processed (because the flag is set to False)

I'll check that reprocessing doesn't raise an error before merge -- that's the only reason I could think to hold off.

nicholas-lockhart (approved):

nicholas-lockhart:
> I think it was originally false simply because we wanted new behavior with the auto process flag to default to existing behavior.

This makes sense.

> It's the cause of a common issue where folks believe their dataset is stuck because the file has been downloaded but not processed (because the flag is set to False)

This also makes sense.

I'm on board!

madisonmay:
It looks like we actually removed the GraphQL call to separately process the files after upload, so setting autoprocess to True is really the only way to add additional files successfully with this version of the client.  

I'll create a ticket to track and then merge.  Thanks folks!

PR #328, opened 2024-08-06
Create inputset python-sdk-publish

indicomachineuser:


.harness/pythonsdkpublish.yaml
```diff
@@ -0,0 +1,89 @@
...
+    stages:
+      - stage:
+          identifier: build_info
+          type: CI
+          spec:
+            infrastructure:
+              type: KubernetesDirect
+              spec:
+                nodeSelector:
+                  node_group: <+input>
```
jacobmanderson, line 24: how do these inputs get populated? it looks like a node taint issue in harness is causing the build to fail

Sung96kim: https://github.com/IndicoDataSolutions/indico-client-python/blob/master-patch-1/.harness/publish_python_sdk.yaml#L30

goatrocks (approved):

